Issue #8, opened 2021-09-14
Required feature flags for example are badly documented


s0lst1ce:
Running the example seems to fail:
```rs
error[E0433]: failed to resolve: could not find `CollectComponentInteraction` in `serenity`
   --> examples/framework_usage/commands.rs:113:37
    |
113 |     while let Some(mci) = serenity::CollectComponentInteraction::new(ctx.discord())
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ could not find `CollectComponentInteraction` in `serenity`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `poise` due to previous error```

kangalioo:
The example requires the collector feature. If you run the example with `cargo run --example=framework_usage --features collector` it should work

s0lst1ce:
Thanks. Should I make a PR to document that?

kangalioo:
Well it is documented in the README though it seems I chose an unfortunate place for it :smile:

> You can run the framework_usage example with cargo run --example=framework_usage --features collector

Feel free to make a PR to make the information more visible

kangalioo:
Thanks for the PR!

PR #9, opened 2021-09-15
add instructions to run example

s0lst1ce:
As mentioned in #8 here's a PR to give instructions as to how to run examples.


s0lst1ce:
I also took the liberty of making the panics more expressive if an env var is missing. Previously all you'd get was an `Error: NotPresent` which you had to guess comes from `std::env::var`.

s0lst1ce:
Of course. Thanks for fixing the typos and rustfmt.

PR #11, opened 2021-10-01
Made prefix dispatch_message public.

vicky5124:
Here's a random example as to why this may be useful.

```rust
/// Runs the command in a message as if you did.
#[poise::command(context_menu_command = "Run for yourself")]
async fn run_yourself(
    ctx: Context<'_>,
    #[description = "Command message to run (enter a link or ID)"] mut msg: Message,
) -> Result<(), Error> {
    msg.author = ctx.author().clone();
    msg.id.0 = ctx.id();

    if let Err(e) = poise::dispatch_message(ctx.framework(), ctx.discord(), &msg, false).await {
        match e {
            Some((err, _ctx)) => {
                poise::say_reply(ctx, "There was an error running the command.").await?;
                return Err(err);
            }
            None => {
                poise::say_reply(ctx, "This message is not a bot command.").await?;
                return Ok(());
            }
        }
    }

    poise::say_reply(ctx, &format!("Successfully ran command: `{}`", msg.content)).await?;

    Ok(())
}
```

	modified:   src/framework/mod.rs

### DIFF
--- a/src/framework/mod.rs
+++ b/src/framework/mod.rs
@@ -10,6 +10,9 @@ pub use builder::*;
 use crate::serenity_prelude as serenity;
 use crate::*;
 
+/// Manually dispatch a message with the prefix framework.
+pub use prefix::dispatch_message;
+
 async fn check_permissions<U, E>(
     ctx: crate::Context<'_, U, E>,
     required_permissions: serenity::Permissions,
--- a/src/framework/mod.rs
+++ b/src/framework/mod.rs
@@ -10,7 +10,6 @@ pub use builder::*;
 use crate::serenity_prelude as serenity;
 use crate::*;
 
-/// Manually dispatch a message with the prefix framework.
 pub use prefix::dispatch_message;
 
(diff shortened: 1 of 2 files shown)


kangalioo:
Thanks for the PR. When making the function public, it would probably be good to exchange the ad-hoc return type into something more understandable like
```rust
struct DispatchResult {
    /// A command was successfully dispatched and run
    CommandSuccess,
    /// No command was run because the message didn't contain a command invocation
    NoCommandRun,
    /// An error occured in user code, for example in the command code or the command check
    Error(E, crate::PrefixCommandErrorContext<'a, U, E>),
}
```
Or is that overkill, what do you think?

vicky5124:
I'm used to `Result<Option<>>` types, with very similar functionality to that, but the enum as a result will allow the user to be able to figure out what everything is without having to look at the documentation of the dispatch method, the only issue I see is that `?` will no longer be able to be used, as the Try trait is nightly.

PR #20, opened 2021-11-01
Typo in example

s0lst1ce:
The first example provided in the README had a syntax typo. Maybe it would be beneficial to make the examples tested to prevent such mistakes in the future? If so I can try to make a PR.

PS: I never had to fetch commits from an upstream repo to a fork so I tried to figure out how to git my out of it. Hopefully I did well.


NotNorom (approved):

kangalioo:
> Maybe it would be beneficial to make the examples tested to prevent such mistakes in the future?

For sure

To make the examples runnable, there would need to be some boilerplate code added. Prefixing those lines of code with `#` hides them in rustdoc _but not_ on GitHub. So if the examples were to be made testable, the GitHub README should probably be almost empty with a link pointing to the docs main page.

Thinking about it, that seems like a good trade-off, though what do you think about it?

s0lst1ce:
I forgot the lines starting with `#` would also show up on GH. The short README is a solution but I personally find working code examples engaging. Since the whole point of GH is displaying your code a well structured, albeit short, README is important to me. It's like the difference between a shop with a vitrine and one without.
Maybe there's a cargo extension that generates a README from the crate-level doc while stripping the `#` in code examples, and maybe resolving doc links. After all making the README, considering there's already the crate-level doc, shouldn't be harder than a simple command. The later could even be done in CI, or at the very least the later should check the command has been ran.
I'll investigate some more and maybe write something if I find the need.

s0lst1ce:
Actually I see the README is not generated from the crate-level doc but it's actually the opposite. Isn't that a missed opportunity?

PR #25, opened 2021-11-11
Fix lifetimes for some `send` functions

snowsee:
Currently, it is not possible to use `CreateReply::attachment` without `'static` `File`s. This is because of an error in the signature of `Context::send` - one of the the `builder` lifetimes should be associated with the `send` function instead of the builder. The same is true for `send_reply`, `send_prefix_reply`, and `send_application_reply` functions.

This PR fixes that by moving the lifetime to the correct position.


kangalioo:
Thank you, nice catch!

PR #32, opened 2021-12-06
Support simd_json

vicky5124:
modified:   src/event.rs

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+rustflags = ["-C", "target-cpu=haswell"]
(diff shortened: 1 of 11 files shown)


vicky5124:
CI is failing, but not from the changes made :D
Fix them, and I'll rebase to make sure.

kangalioo:
There's a way that doesn't require specialized code for serde_json and simd_json, so I took the liberty of making that change (and rebasing), hope you don't mind.

Also, what does that `.cargo/config.toml` file do?

vicky5124:
`.cargo/config.toml` is there to set `RUSTFLAGS` on the cargo commands ran on the project.
The flag that's set is required for simd_json to compile, as it won't compile unless the CPU supports the SSE4.2 or AVX2 instruction sets on x86_64, or NEON on Arm; it tells the compiler to use the intel Haswell architecture as a minimum supported so the CI can compile the code because the compiler can now use AVX2. If that makes sense.

PR #34, opened 2021-12-10
Fixed matching subcommands with the same name

noituri:
Hi,
I've had a problem with two slash commands `/world info` and `/plot info`. Both `info`s are subcommands. I fixed it by comparing interaction name with slash command meta name


kangalioo:
Thank you!

PR #41, opened 2022-02-01
Ensure global checks are always ran

JellyWX:
Fixes issue where if a command specified its own checks, global checks (those registered on the framework) would not be ran


kangalioo:
Thanks!

PR #45, opened 2022-02-20
change identifying name

JellyWX:
Proposed change that sets the identifying name to default to the function identifier rather than the command name. This change helps keep identifying names unique even with subcommands.

An alternative could be to create the identifying name from the full qualified command name, but this was a quicker potential solution.

Change could be breaking.


kangalioo:
@vicky5124 Is this breaking change ok with you?

vicky5124:
I'm ok with the command function name, as for my use case, the latter is less convenient. OFC it's still configurable, so it's good anyways.

PR #46, opened 2022-02-20
Dont send defers twice

JellyWX:
Don't send defers if an interaction response has already been sent


kangalioo:
Makes sense.

~~It might be better to replace the two AtomicBool calls with a single `self.has_sent_initial_response.swap(true, std::sync::atomic::Ordering::SeqCst)` check? It checks the value and sets it to true at the same time~~ Nevermind, the HTTP request can fail so the current version is more correct.

PR #53, opened 2022-04-01
Add ctx and framework to listener error

GnomedDev:
`FrameworkError::Listener` gets a surprising lack of `serenity::Context` or `Framework`, let's fix that!


kangalioo:
Can you run cargo fmt?

PR #55, opened 2022-04-02
Fix serenity version

Gabriel-Paulucci:
I removed the cargo lock because the cargo lock is not something that should be in the lib repository, and that caused the serenity version to not be fixed in commit #54


kangalioo:
I deliberately added Cargo.lock in 8bcafc112231d3b7df75bf10e48980c3fa724370 to make tests and examples deterministic. Cargo.lock in a library does no harm because it is ignored when the library is used as a dependency

The CI breakage was unrelated to #54; it happened because upstream serenity rebased the next branch, which changed commit hashes and invalidated the stored serenity commit hash (see my fix commit 953f863992ef9f49889be91e108f24ef4a9d1873)

Said commit 953f863992ef9f49889be91e108f24ef4a9d1873 also updates the serenity hash and fixes the CI breakage, so I think this PR is not needed anymore? Thank you in any case for contributing a fix!

Gabriel-Paulucci:
Oh yes I'm sorry I didn't know it was working like this, I will close the PR

PR #59, opened 2022-04-08
Add more command checks

devtomio:
- `guild_only` - only run on guilds
- `dm_only` - only run on dm channels (opposite of `guild_only`)
- `nsfw_only` - only run on nsfw channels

### DIFF
--- a/src/framework/dispatch/common.rs
+++ b/src/framework/dispatch/common.rs
@@ -79,15 +79,16 @@ pub async fn check_permissions_and_cooldown<'a, U, E>(
     }
 
     if cmd.nsfw_only {
-        match ctx.guild() {
-            Some(guild) => match guild.nsfw_level {
-                serenity::NsfwLevel::AgeRestricted | serenity::NsfwLevel::Explicit => {
-                    return Err(crate::FrameworkError::NsfwOnly { ctx })
-                }
-                _ => (),
-            },
-            None => return Err(crate::FrameworkError::NsfwOnly { ctx }),
-        };
+        match ctx
+            .channel_id()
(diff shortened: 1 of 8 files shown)


src/framework/dispatch/common.rs
```diff
@@ -70,6 +70,27 @@ pub async fn check_permissions_and_cooldown<'a, U, E>(
...
+    if cmd.dm_only && ctx.guild_id().is_some() {
+        return Err(crate::FrameworkError::DmOnly { ctx });
+    }
+
+    if cmd.nsfw_only {
+        match ctx
+            .channel_id()
+            .to_channel(ctx.discord())
+            .await
+            .unwrap()
```
kangalioo, line 86: Can you replace the unwrap? In case of an error, you can log a warning with `log::warn!` and return NsfwOnly error just to be safe

kangalioo (changes requested):
Seems good for the most part, thanks!

src/framework/dispatch/common.rs
```diff
@@ -70,6 +70,29 @@ pub async fn check_permissions_and_cooldown<'a, U, E>(
...
+        let channel = match ctx.channel_id().to_channel(ctx.discord()).await {
+            Ok(channel) => channel,
+            Err(e) => {
+                log::warn!("Error when getting channel: {}", e);
+
+                return Err(crate::FrameworkError::NsfwOnly { ctx });
+            }
+        };
+
+        if channel.is_nsfw() {
```
kangalioo, line 91: Add `!` ;)

devtomio: great catch, missed it lol

PR #66, opened 2022-04-26
Use serenity's support for adding attachments to initial responses

GnomedDev:
Probably a breaking change, lifetimes are probably horrible


kangalio:
Blocked on https://github.com/serenity-rs/serenity/pull/1874 releasing to crates.io

PR #69, opened 2022-05-04
Made the changes to the register_application_commands method.

Bloectasy:


kangalio:
I took the liberty of changing the layout a bit

Before:
Screenshot_20220507_020618

After:
Screenshot_20220507_020640

(blue is for call to action, new wording seems less clunky, 2x2 layout fits closer to the buttons purposes)

PR #70, opened 2022-05-04
permissions v2

JellyWX:
targeting develop as this relies on git serenity. basic support for permissions v2 as a new `default_member_permissions` field on the command macro.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -971,8 +971,9 @@ dependencies = [
 
 [[package]]
 name = "serenity"
-version = "0.11.1"
-source = "git+https://github.com/serenity-rs/serenity?branch=current#e690ccd17174f2b9dc3f6a1545f69a209a11241f"
+version = "0.11.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "246008828de84241202b092316c4729da36c3edd8f0c17ed6893df8e414c3c9b"
 dependencies = [
  "async-trait",
  "async-tungstenite",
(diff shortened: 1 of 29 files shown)


kangalio:
Thank you!

Why didn't you adapt the existing required_permissions for this?

JellyWX:
I wasn't certain about the idea of having what would essentially be two separate permission check systems under the same name, as prefix commands would still check internally with the bot and slash commands would no longer.

If you're okay with that I'll change it. I don't use prefix commands at all so it doesn't really affect me

kangalio:
Okay cool, yeah I would prefer reusing required_permissions for simplicity, and to avoid existing users having to port to the new field

The fact that Discord permissions v2 work slightly differently, (guild owners can modify permission checks if they want (?)) is fine I think

JellyWX:
The issue here is with subcommands, since those don't use permissions properly I don't think (on permissions v2). Also, I haven't been able to test this commit properly yet due to https://github.com/discord/discord-api-docs/issues/4856

kangalio:
By now I feel like it's better to have a separate attribute argument `default_member_permissions` after all. That way, users explicitly choose to opt in to the subtle changes: guild owners being able to change permissions; no subcommand specific permissions

kangalio:
@jellywx Are you ok with the PR being merged like this?

JellyWX:
Yep,looks fine by me. Thanks!

Sent from ProtonMail mobile

-------- Original Message --------

PR #71, opened 2022-05-05
Minimize monomorphization-caused codegen bloat

elkowar:
the send_application_reply and send_prefix_reply functions are quite long, very commonly called functions, which are generic over their respective builder function. 
During monomorphization, these function bodies get duplicated for each different closure passed to the functions, thus causing a lot of unnecessary codegen output.
By putting most of the function body into a non-generic inner function, we can minimize the impact of this, resulting in a noticable decrease in generated LLVM.

### DIFF
--- a/src/prefix/track_edits.rs
+++ b/src/prefix/track_edits.rs
@@ -164,66 +164,67 @@ pub async fn send_prefix_reply<'a, U, E>(
     ctx: crate::prefix::PrefixContext<'_, U, E>,
     builder: impl for<'b> FnOnce(&'b mut crate::CreateReply<'a>) -> &'b mut crate::CreateReply<'a>,
 ) -> Result<Box<serenity::Message>, serenity::Error> {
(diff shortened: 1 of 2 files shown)


kangalio:
Great, thank you!

Do you happen to know how much this change impacts codegen sizes or compile times exactly? Out of curiosity

Also, could you make the inner functions into private outer functions instead, to keep the indentation the same and make the git diff cleaner?

elkowar:
Done!
Well, as mentioned in discord, compile times are very hard to meaningfully analyze here, sadly.
In my project, this change resulted in an overall llvm-lines reduction of about 2%.
Maybe a bit more meaningfully:
If I'm not very bad at interpreting the data (or at math, I guess), then this brings the LLVM-lines per copy of
- `send_prefix_reply` down from **916** to **292**
- `send_application_reply` down from **678** to **292**
(that's simply me dividing the total lines caused by the functions by the number of copies throughout my codebase)

PR #74, opened 2022-05-09
Add intercept_dispatch to FrameworkOptions

elkowar:
This PR adds a new option to the FrameworkOptions struct: `intercept_dispatch`.
This allows users to wrap the entire event-dispatch logic in their own code.
For my project, this is required to allow us to properly provide contextual log-traces.
Additionally, this could be useful to add additional checks and profiling or debugging-related logic to the event dispatch system,
as well as possibly making some other operations possible that would previously have depended on adding more specific features to poise.

### DIFF
--- a/src/framework/mod.rs
+++ b/src/framework/mod.rs
@@ -109,10 +109,20 @@ impl<U, E> Framework<U, E> {
             let existing_event_handler = existing_event_handler.clone();
 
             Box::pin(async move {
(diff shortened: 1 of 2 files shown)


elkowar:
looking at the failing tests...

elkowar:
I'd say we can close this, given that you implemented a5ac0da07e490f613f7770848819f5dd1755e479, which can fully replace this

kangalio:
Great

Thank you for your effort anyways!

PR #75, opened 2022-05-11
Add set_components, set_embed and add_embed methods to CreateReply

elkowar:
Serenity provides builder methods that take owned `CreateEmbed` and `CreateComponents` objects in addition to the lambda-based builder methods. This PR adds equivalent methods to poise's CreateReply object, for the sake of convenience and consistency

### DIFF
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -44,6 +44,23 @@ impl<'att> CreateReply<'att> {
         self
     }
 
+    /// Adds an embed to the message.
+    ///
+    /// Existing embeds are kept.
+    pub fn add_embed(&mut self, embed: serenity::CreateEmbed) -> &mut Self {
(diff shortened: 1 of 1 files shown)


kangalio:
I deliberately avoided these methods because they duplicate functionality for (imo) little gain. You can already do `builder.embed(|b| {*b = embed; b})` or `builder.embeds.push(embed)`

elkowar:
Ohh, I didn't realize that the fields where public -- I guess that fills the same usecase, yea!
I'd still think being consistent with the serenity-api for very little cost might be worth it, but If you'd rather keep it minimal then feel free to close this 👍🏻 Understandable either way

kangalio:
Thank you for your understanding 😊

PR #78, opened 2022-05-28
Fix clippy warnings

GnomedDev:
Does what it says on the tin!

### DIFF
--- a/examples/quickstart/main.rs
+++ b/examples/quickstart/main.rs
@@ -11,7 +11,7 @@ async fn age(
     ctx: Context<'_>,
     #[description = "Selected user"] user: Option<serenity::User>,
 ) -> Result<(), Error> {
-    let u = user.as_ref().unwrap_or(ctx.author());
+    let u = user.as_ref().unwrap_or_else(|| ctx.author());
     let response = format!("{}'s account was created at {}", u.name, u.created_at());
     ctx.say(response).await?;
     Ok(())
--- a/src/dispatch/common.rs
+++ b/src/dispatch/common.rs
@@ -76,9 +76,10 @@ async fn missing_permissions<U, E>(
 /// Doesn't actually start the cooldown timer! This should be done by the caller later, after
 /// argument parsing.
 /// (A command that didn't even get past argument parsing shouldn't trigger cooldowns)
+#[allow(clippy::needless_lifetimes)] // false positive (clippy issue 7271)
 pub async fn check_permissions_and_cooldown<'a, U, E>(
     ctx: crate::Context<'a, U, E>,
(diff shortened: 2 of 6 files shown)


examples/quickstart/main.rs
```diff
@@ -11,7 +11,7 @@ async fn age(
     ctx: Context<'_>,
     #[description = "Selected user"] user: Option<serenity::User>,
 ) -> Result<(), Error> {
-    let u = user.as_ref().unwrap_or(ctx.author());
```
kangalio, line 14: Is that a default clippy lint? Clippy doesn't complain when I run it locally (latest nightly)

src/dispatch/common.rs
```diff
@@ -76,10 +76,9 @@ async fn missing_permissions<U, E>(
 /// Doesn't actually start the cooldown timer! This should be done by the caller later, after
 /// argument parsing.
 /// (A command that didn't even get past argument parsing shouldn't trigger cooldowns)
-#[allow(clippy::needless_lifetimes)] // false positive (clippy issue 7271)
 pub async fn check_permissions_and_cooldown<'a, U, E>(
     ctx: crate::Context<'a, U, E>,
-    cmd: &crate::Command<U, E>,
+    cmd: &'a crate::Command<U, E>,
```
kangalio: This is not equivalent to before. Like this, check_permissions_and_cooldown can't accept commands with a lifetime smaller than the Context lifetime anymore. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=349063a911b6803bf8d2b5d1a46f8351

kangalio:
Could you
- remove the lifetime change
- remove the newlines between imports (for consistency)
- replace `writeln!().unwrap()` with `let _ = writeln!()` (to avoid meaningless unwraps)

GnomedDev:
On the removing `unwrap`s, that would trigger a pedantic warning, and a long term goal of mine is to be bored enough to make poise pass pedantic, can I keep as is (in this PR)?

kangalio:
Which lint is triggered?

In general I'm not a fan of clippy::pedantic because almost all changes are noise at that point

GnomedDev:
clippy::let_underscore_drop, I had a lot of fun fixing it in serenity /s

kangalio:
Hmm

> let_underscore_drop
> Why is this bad?
> If you want to explicitly drop the expression, std::mem::drop conveys your intention better and is less error-prone.

<https://rust-lang.github.io/rust-clippy/master/#let_underscore_drop>

I wouldn't want to throw `std::mem::drop(write!())` over `let _ = write!()` everywhere tbh. This would be one of the lints to `#![allow()]` imo

PR #80, opened 2022-06-01
Fix default_member_permissions

GnomedDev:
Permissions::empty means only admins can run the command, not that there are no permission requirements. Only not supplying the field allows anyone to use the command.


kangalio:
Context on my commit: https://discord.com/channels/381880193251409931/919310428344029265/981550285363249203

PR #82, opened 2022-06-11
Localization

### LINKED ISSUE #79: localization
when https://github.com/serenity-rs/serenity/pull/1921 gets merged, how will this library handle localizations?

kangalio:
Fixes #79

As described in #79, there are two ways to localize commands: via attributes, or programmatically.

Attributes are good for small scale bots because they're easy to get started with and simple: for example `#[name_localized("de", "Befehlsname")]` to set the German-localized name to "Befehlsname". A full example is in examples/framework_usage/commands/localization.rs

The programmatical approach is better for larger bots that use a translation framework like gettext, fluent: you can insert localizations of all commands with a single code snippet by modifying the `name_localizations` and `description_localizations` fields of `Command`, `CommandParameter` and `CommandParameterChoice`. A full example is in examples/fluent_localization/

CI on this PR currently fails due to deprecation warnings that haven't been released to crates.io. I'd like to fix those warnings in a separate commit, so I'll leave this PR's CI failing.


kangalio:
I still need to fix the fluent_localization example to fall back on en-US for unsupported locales instead of the raw key

Also, the entire localization system is slash-only but I think that's fine

kangalio:
Main problem left now is GNU gettext: its peculiar way of gathering translation keys by searching the source code for `gettext("...")` invocations is incompatible with poise's command macro: you can't write

```rust
#[poise::command]
fn gettext("command_name")(
    gettext("param1"): String,
    gettext("param2"): String,
) {}
```

The workaround would be:

```rust
#[poise::command(rename = gettext("command_name")]
fn command_name(
    #[rename = gettext("param1")] param1: String,
    #[rename = gettext("param2")] param2: String,
) {}
```

_However_ this is hella noisy plus it's completely incompatible with the entire current attribute parsing infrastructure. Currently, poise parses attributes via darling, which uses syn's Meta structs, but that is really crippled and only supports very limited syntax - not including key-value params with non-literal values like `rename = gettext("...")` ([values are hardcoded to literals](https://docs.rs/syn/latest/syn/struct.MetaNameValue.html)). We'd have to replace the parsing infrastructure with something TokenStream-based: hell of a lot of work.

Soo I think I'm gonna not do that for now and keep this PR as is. The new parsing infrastructure can be added later in a backwards compatible way

kangalio:
This PR depends on serenity's localization support and will therefore be merged once serenity 0.11.3 releases

PR #85, opened 2022-06-19
Don't keep edit tracker task around if no edit tracker

GnomedDev:
This saves a background task if the bot isn't actually using edit tracking


kangalio:
Hmm I wonder what my original thought process was there. I might have wanted to prepare for dynamically adding or removing an edit tracker at run time...? Not like that'd be possible

PR #90, opened 2022-06-28
Fix deprecation warnings for `serenity::interactions`

nickelc:
This fixes the deprecation warnings for the upcoming serenity 0.11.3 version.


src/structs/slash.rs
```diff
@@ -18,7 +18,7 @@ pub enum ApplicationCommandOrAutocompleteInteraction<'a> {
 
 impl<'a> ApplicationCommandOrAutocompleteInteraction<'a> {
```
kangalio, line 19: Can be renamed to CommandOrAutocompleteInteraction

examples/framework_usage/commands/general.rs
```diff
@@ -133,7 +133,7 @@ pub async fn boop(ctx: Context<'_>) -> Result<(), Error> {
         .await?;
 
         mci.create_interaction_response(ctx.discord(), |ir| {
-            ir.kind(serenity::InteractionResponseType::DeferredUpdateMessage)
+            ir.kind(serenity::interaction::InteractionResponseType::DeferredUpdateMessage)
```
kangalio: Why not keep `serenity::InteractionResponseType` as before?

nickelc: The use of `serenity::prelude::*` in `poise`'s prelude includes the deprecated variant from `serenity::interactions` and this leads to:
```
error[E0659]: `InteractionResponseType` is ambiguous
   --> src/modal.rs:64:30
    |
64  |             b.kind(serenity::InteractionResponseType::DeferredUpdateMessage)
    |                              ^^^^^^^^^^^^^^^^^^^^^^^ ambiguous name
    |
    = note: ambiguous because of multiple glob imports of a name in the same module
note: `InteractionResponseType` could refer to the enum imported here
   --> src/lib.rs:268:94
    |
268 |                     application_command::*, autocomplete::*, message_component::*, modal::*, *,
    |                                                                                              ^
    = help: consider adding an explicit import of `InteractionResponseType` to disambiguate
note: `InteractionResponseType` could also refer to the type alias imported here
   --> src/lib.rs:272:13
    |
272 |             prelude::*,
    |             ^^^^^^^^^^
    = help: consider adding an explicit import of `InteractionResponseType` to disambiguate
```

src/structs/slash.rs
nickelc: sure, but I think that's out of scope for the PR.
I also have some changes in the pipeline that will lead to more changes like this.

<details>
<summary>
<b>Merge application command & autocomplete interactions</b>

    The structs are almost identical except for the `focused` field for the
    current autocompleted option and that the `resolved` data map is not
    sent for autocomplete interactions.
    ...
</summary>

`autocomplete.rs` is gone.
```
 examples/e14_slash_commands/src/main.rs                  |  34 ++---
 src/model/application/interaction/application_command.rs | 720 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--------
 src/model/application/interaction/autocomplete.rs        | 516 ---------------------------------------------------------------------------
 src/model/application/interaction/mod.rs                 |  23 ++--
 src/model/utils.rs                                       |  88 -------------
```
</details>

examples/framework_usage/commands/general.rs
kangalio: Weird; when I remove the `interaction::`, it just errors:

```rust
error[E0433]: failed to resolve: could not find `InteractionResponseType` in `serenity`
   --> examples/framework_usage/commands/general.rs:136:31
    |
136 |             ir.kind(serenity::InteractionResponseType::DeferredUpdateMessage)
    |                               ^^^^^^^^^^^^^^^^^^^^^^^ could not find `InteractionResponseType` in `serenity`
```

Adding `pub use serenity::model::application::interaction::InteractionResponseType;` into serenity_prelude fixes all errors.

Am I doing something wrong?? Any idea why I'm not seeing the error you're seeing (which is making much more sense than mine)

src/structs/slash.rs
kangalio: Are you implying you will merge ApplicationCommandInteraction and AutocompleteInteraction in upstream serenity soon? Interesting

examples/framework_usage/commands/general.rs
nickelc: I didn't know this error but I get the same error in `general.rs`.
My error message is from `modal.rs`

src/structs/slash.rs
nickelc: Yes, but I want to check out how this will improve poise or not.

I removed the cloning of the entites from the `resolved` data map and replaced it with
a method `CommandData::options() -> Vec<ResolvedOption<'_>>` returning references.
 
```rust
impl CommandData {
    pub fn autocomplete(&self) -> Option<AutocompleteOption<'_>> {
    }
    pub fn options(&self) -> Vec<ResolvedOption<'_>> {
    }
}

pub struct AutocompleteOption<'a> {
    pub name: &'a str,
    pub kind: CommandOptionType,
    pub value: &'a str,
}
```

src/lib.rs
```diff
@@ -250,6 +250,7 @@ pub use serenity;
 /// use poise::serenity_prelude as serenity;
 /// ```
 pub mod serenity_prelude {
+    pub use serenity::model::application::interaction::{Interaction, InteractionResponseType};
```
kangalio: Can you add the other types that have had same-named type aliases added, too? Those are, in addition to Interaction and InteractionResponseType: InteractionApplicationCommandCallbackDataFlags, InteractionType, MessageInteraction

https://github.com/serenity-rs/serenity/pull/1899/files#diff-50cf0c1ab14aa2f5fce5fe01f9f241f8131c383e8973875b2fc0f7e495f57f1cR170-R189

kangalio: Btw since this is a draft PR; would it better if I refrain from reviewing until the PR is done?

nickelc: `interaction::MessageFlags` would probably overload `channel::MessageFlags`.

The PR is done beside the prelude additions.

nickelc: I could add `interaction::MessageFlags` as `InteractionMessageFlags`.

kangalio (approved):

kangalio:
As with #82, I'll merge this PR once the next serenity version releases

src/lib.rs
kangalio: I already added it as InteractionResponseFlags in the most recent commit. You can rename to InteractionMessageFlags though if you prefer that name

PR #92, opened 2022-07-04
Prepare poise for serenity's `next` branch

nickelc:
The changes are based on a slightly modified [`next`] branch of serenity to track its breaking change in chronological order and making it easier for the upcoming merge of the application command & autocomplete types.

The poise & examples compile but are currently untested.

I left some TODOs in the "new builder" commit because of missing functionality.

[`next`]: https://github.com/nickelc/serenity/tree/staging/poise-next


kangalio:
Author stated they are no longer working on the PR so I'll close this

Serenity next branch changes so much it's probably easier to just re do the PR when 0.12 releases instead of keeping up with constant changes that are also sometimes reverted

PR #95, opened 2022-07-26
Enable manual sharding configurations using `Framework::start_with`

ChargedByte:
Adds methods to start the framework with different sharding configurations available in the `serenity::Client`

### DIFF
--- a/src/framework/mod.rs
+++ b/src/framework/mod.rs
@@ -1,9 +1,10 @@
 //! The central Framework struct that ties everything together.
 
-mod builder;
 pub use builder::*;
 
 use crate::{serenity_prelude as serenity, BoxFuture};
+
+mod builder;
 
 /// The main framework struct which stores all data and handles message and interaction dispatch.
 ///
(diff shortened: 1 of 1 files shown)


kangalio:
Thanks

Maybe the start_with method should be exposed instead so we don't have to add so many boilerplatey wrapper functions? `.start()` and `.start_autosharded()` are used the most (I believe?) so those wrapper functions could stay for convenience

ChargedByte:
Sure, that makes sense. The `Framework::start_with` method is already public but with the original `fn(Client) -> F` type, trying to move variables into closure causes an error 'closures can only be coerced to `fn` types if they do not capture any variables'. The first commit in the chain ([eb6b151](https://github.com/kangalioo/poise/pull/95/commits/eb6b151271c08da678af2a91e7295bd65113b8d3)) is enough to fix this.

kangalio:
> The Framework::start_with method is already public

Oops I forgot lol

Okay there should be a note in the start() and start_autosharded() docs explaining the use of start_with for serenity's other start functions. You can do it or I will once I'm home from vacation

ChargedByte:
I can do it while yanking those methods I added.

ChargedByte:
That should do it?

I only added a single example of how to use the `Framework::start_with` method but I think it's enough to demonstrate how it works.

firefox_7ZubmcEmfR

firefox_s7gYAyR61y

kangalio:
Are you ok with my changes?

ChargedByte:
LGTM

PR #96, opened 2022-07-26
Rename Framework build method to builder

ChargedByte:
Deprecate `Framework::build` in favour of `Framework::builder`, this would be more in line with the common standard (at least from what I've seen).


kangalio:
Thanks, yes this seems like the more common standard; looking around. And you even left a deprecated alias, great!

PR #105, opened 2022-08-31
Revert "fix: to_string in command macro clashes with lint clippy::str_to_string"

mrnossiom:
Reverts kangalioo/poise#104


mrnossiom:
I didn't investigate enough, it also can happen in the parameters name.

mrnossiom:
I've added a `#[allow(clippy::str_to_string)]` attribute to the command macro generation

PR #107, opened 2022-09-06
add poise::Context::category_id

Monadic-Cat:
It'd be nice if we could easily fetch the containing category for the channel a command is invoked in.
This method tolerates the Serenity cache not being populated (and thus has to be `async` in order to fetch that info from Discord).

### DIFF
--- a/src/structs/context.rs
+++ b/src/structs/context.rs
@@ -136,16 +136,16 @@ impl<'a, U, E> Context<'a, U, E> {
     }
 
     /// Return the category ID of this context, if we are inside a category
+    ///
+    /// If we're not in a category channel, or not even in a guild, Ok(None) is returned.
+    /// If retrieving channel data failed, Err(...) is returned.
     pub async fn category_id(&self) -> Result<Option<serenity::ChannelId>, serenity::Error> {
-        let chan = self.discord().cache.channel(self.channel_id());
-        let chan = match chan {
-            Some(chan) => chan,
-            None => self.discord().http.get_channel(self.channel_id().0).await?,
-        };
-        Ok(match chan {
-            serenity::Channel::Guild(chan) => chan.parent_id,
-            _ => None
-        })
+        Ok(self
+            .channel_id()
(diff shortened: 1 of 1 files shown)


kangalio:
I changed the implementation to use channel_id.to_channel(), which already implements the cache fallback on HTTP error mechanism (plus it also supports the new temp_cache feature). This also means the function now compiles without cache feature enabled (and it has slightly fewer lines of code). I also added some more documentation.

But looking at this, this function is basically just a wrapper around `to_channel().guild().parent_id`, and there's loads other fields on GuildChannel that should have such functions, if there's one for parent_id. So if anything there should probably just be a function for `to_channel().guild()` and the user can access the field themselves. However there's also other channel types that could warrant such a method. Following that train of reasoning, I'm not sure if any new method should be added at all, to be honest...?

Monadic-Cat:
Dunno. I wasn't aware that `to_channel()` existed, and I had to go back and read Discord docs to discover "oh it's probably not exposed here because an Interaction object doesn't carry that field", and then go figure out where to grab the category ID from instead.
I expected this method to be here initially because both `channel_id()` and `guild_id()` are here, but if that's not compelling, I'm not super attached to this being merged, since it's just a wrapper for existing public API.

kangalio:
> I wasn't aware that to_channel() existed, and I had to go back and read Discord docs to discover "oh it's probably not exposed here because an Interaction object doesn't carry that field", and then go figure out where to grab the category ID from instead.

Sounds like I should improve documentation. I tried https://docs.rs/poise/latest/poise/#finding-serenity-methods to point people to the very useful methods on the Id types but that section is evidently not visible or understandable enough. What do you think would help the docs so that you wouldn't have had such a hard time discovering how to get category_id?

kangalio:
Added more docs in 5ba00794e65766a9c4d72a1560043940210b9f89

https://kangalioo.github.io/poise/develop/poise/index.html#useful-serenity-methods

PR #112, opened 2022-10-09
Add dm_permissions as an available option to application commands part 2

Bloectasy:
Forgot to add the dm_only part.

### DIFF
--- a/src/structs/command.rs
+++ b/src/structs/command.rs
@@ -191,6 +191,8 @@ impl<U, E> Command<U, E> {
 
         if self.guild_only {
             builder.dm_permission(false);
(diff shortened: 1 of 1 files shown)


kangalio:
`.dm_permission(true)` enables the command for DMs, but doesn't disable them in guilds.

There's no designated attribute to disable commands in guilds, so our only option is `.default_permission(false)` which is deprecated and is overridable by guild owners, or `.default_member_permissions(all permissions)` which doesn't keep guild admins from seeing the command and it's also overridable by guild owners.

Seeing as Discord doesn't officially have built-in support for dm_only and the workarounds don't work well, I think we shouldn't implement this

PR #114, opened 2022-10-13
Update serenity

jepcd:
https://discord.com/channels/381880193251409931/919310428344029265/1030184268309200949

PR #116, opened 2022-10-18
Fix serenity-next branch

GnomedDev:
Fixes compilation with the serenity next branch by:
- Using `as_deref` to fix the box change
- Swapping `CacheAndHttp` for `Context` in prefix argument parsing
- Fixing the event macro to match serenity by `&GuildChannel` -> `GuildChannel` (gets rid of the lifetime aswell)


kangalio:
Could you fix the errors? (wrong timestamp method?) Not sure what the MSRV is on about, maybe it's fixed if we just raise poise MSRV on serenity-next to serenity's MSRV on next

kangalio:
Closed in favor of #114. Thanks though :)

PR #117, opened 2022-10-18
Generate unique IDs for modals

p5nbTgip0r:
## Issue
When a modal is created using `Modal::execute` or `execute_with_defaults`, the user can close the modal prompt, re-execute the command to receive another modal, and finally submit the modal. When this happens, poise will attempt to respond to the interaction twice.

## Reproduction
1. [Using the example modal code](https://github.com/kangalioo/poise/blob/162ac53094d1daec4e96f67980aaf50823e6f46a/examples/framework_usage/commands/general.rs#L280-L296), register the modal command and execute it
2. Close the modal without submitting (`X` in the top right of the modal, `Cancel`, press ESC, back button, etc.)
3. Execute the modal command again
4. Submit the modal using the `Submit` button
5. Discord will give an error on the HTTP request stating either `Unknown interaction` or `Interaction has already been acknowledged.`

## Cause
[At this point](https://github.com/kangalioo/poise/blob/5ba00794e65766a9c4d72a1560043940210b9f89/src/modal.rs#L56) a `CollectModalInteraction` collector is created which listens for modal submissions by the author. If the author closes the modal, re-opens it (thus creating another collector), and submits it, both collectors will be fulfilled and the interaction will be responded to twice.

## Solution
The solution I came up with is to set the modal's `custom_id` to the initial interaction ID, and then filter for modal submission events with the same `custom_id`. This makes it so each modal interaction collector will only collect the modal generated by the initial interaction, and not modals created from other interaction events.

This solution also solves an issue where defaults provided via `execute_with_defaults` may not be set on the client side if the user closes and re-opens the same modal.

### DIFF
--- a/src/modal.rs
+++ b/src/modal.rs
@@ -46,7 +46,7 @@ async fn execute<U: Send + Sync, E, M: Modal>(
     // Send modal
     interaction
         .create_interaction_response(ctx.discord, |b| {
-            *b = M::create(defaults, interaction.id.to_string());
+            *b = M::create(defaults, interaction_id.clone());
             b
         })
         .await?;
--- a/src/modal.rs
+++ b/src/modal.rs
@@ -109,7 +109,10 @@ pub trait Modal: Sized {
     ///
     /// Optionally takes an initialized instance as pre-filled values of this modal (see
     /// [`Self::execute_with_defaults()`] for more info)
-    fn create(defaults: Option<Self>, custom_id: Option<String>) -> serenity::CreateInteractionResponse<'static>;
+    fn create(
+        defaults: Option<Self>,
+        custom_id: String,
+    ) -> serenity::CreateInteractionResponse<'static>;
 
     /// Parses a received modal submit interaction into this type
     ///
--- a/src/modal.rs
+++ b/src/modal.rs
@@ -41,11 +41,12 @@ async fn execute<U: Send + Sync, E, M: Modal>(
     defaults: Option<M>,
 ) -> Result<M, serenity::Error> {
     let interaction = ctx.interaction.unwrap();
+    let interaction_id = interaction.id.to_string();
 
     // Send modal
     interaction
         .create_interaction_response(ctx.discord, |b| {
-            *b = M::create(defaults);
+            *b = M::create(defaults, Some(interaction_id.clone()));
             b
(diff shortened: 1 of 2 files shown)


src/modal.rs
```diff
@@ -108,7 +109,7 @@ pub trait Modal: Sized {
     ///
     /// Optionally takes an initialized instance as pre-filled values of this modal (see
     /// [`Self::execute_with_defaults()`] for more info)
-    fn create(defaults: Option<Self>) -> serenity::CreateInteractionResponse<'static>;
+    fn create(defaults: Option<Self>, custom_id: Option<String>) -> serenity::CreateInteractionResponse<'static>;
```
kangalio: Can you remove the Option from the custom_id parameter? I think it's confusing to pretend you can omit the custom_id, because the implementation will unconditionally set a custom_id anyways

src/modal.rs
```diff
@@ -41,11 +41,12 @@ async fn execute<U: Send + Sync, E, M: Modal>(
     defaults: Option<M>,
 ) -> Result<M, serenity::Error> {
     let interaction = ctx.interaction.unwrap();
+    let interaction_id = interaction.id.to_string();
 
     // Send modal
     interaction
         .create_interaction_response(ctx.discord, |b| {
-            *b = M::create(defaults);
+            *b = M::create(defaults, Some(interaction_id.clone()));
```
kangalio: Can't you just pass `interaction.id.to_string()` directly in here?

kangalio:
Thank you for the PR, good catch! I left a couple comments. (Can't make those tweaks myself since I only have a mobile phone right now)

src/modal.rs
p5nbTgip0r: Yeah, I wasn't sure whether to do that or just re-use the existing variable I made. I've changed it to use `interaction.id.to_string()` though.

kangalio: Oh, I didn't realize `interaction.id.to_string()` was used later again. Then it makes total sense to use a variable instead of inlining. I'll quickly change it back locally so you don't have to go through the trouble

kangalio:
Thank you!

p5nbTgip0r:
Thank you for making this crate!

PR #118, opened 2022-10-21
Enhancements for the register button command(s)

SticksDev:
In this PR, I add:

- All buttons now have emojis, 📋for register global/guild and 🗑️ for delete global/guild.
- Reworded the timeout message
- Add emojis to messages where it seemed appropriate. 
- Message also changes to a "processing" message.
- Adds a time took at the end

Screenshots:
image
image

### DIFF
--- a/src/builtins/register.rs
+++ b/src/builtins/register.rs
@@ -1,4 +1,5 @@
 //! Utilities for registering application commands
+
 use crate::serenity_prelude as serenity;
 
 /// Collects all commands into a [`serenity::CreateApplicationCommands`] builder, which can be used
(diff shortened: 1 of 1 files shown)


src/builtins/register.rs
```diff
@@ -153,23 +152,27 @@ pub async fn register_application_commands_buttons<U, E>(
                             b.custom_id("register.guild")
                                 .label("Register in guild")
                                 .style(serenity::ButtonStyle::Primary)
+                                .emoji(serenity::ReactionType::Unicode("📋".to_string()))
```
kangalio: `.emoji()` takes `impl Into<ReactionType>`, so any type that implements the `Into<ReactionType>` trait. As you can see [here](https://docs.rs/serenity/latest/serenity/model/channel/enum.ReactionType.html#impl-From%3Cchar%3E-for-ReactionType), this includes `char`. So you can just write `.emoji('📋')`

src/builtins/register.rs
```diff
@@ -184,10 +187,12 @@ pub async fn register_application_commands_buttons<U, E>(
         .await;
 
     reply.edit(ctx, |b| b.components(|b| b)).await?; // remove buttons after button press
+    // NOTE: Can this be done in one .edit?
+    reply.edit(ctx, |b| b.content("Processing... Please wait.")).await?; // Edit message to processing after button press
```
kangalio: This should be possible to do in one edit, if you just chain the builder function calls after another `.components(...).content(...)`

src/builtins/register.rs
```diff
@@ -1,5 +1,4 @@
 //! Utilities for registering application commands
-
 use crate::serenity_prelude as serenity;
```
kangalio: Tiny thing but could you add back the newline :innocent: All files have space between the module docs and the other items, and I think it looks neater (I personally separate each group of items at the top of a file with a newline)

SticksDev:
Should be all Resloved for you

kangalio:
Thank you!

PR #122, opened 2022-11-02
Update serenity

jepcd:
- Supports multiple event handlers
- Fixes example

The derive macro poise::Modal seems to be broken, but idk how to fix it


jepcd (commented):
`this function takes 2 arguments but 0 arguments were supplied
two arguments are missing`

kangalio:
Thanks!

PR #123, opened 2022-11-03
Expose Framework.bot_id in the same way as user_data

GnomedDev:
Exposes the `bot_id` field stored in Framework using the same busy loop as `user_data`. This publicly exposes constructing `FrameworkContext` with only a `&Framework` and without repeating the logic poise already does internally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,9 +28,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.57"
+version = "0.1.58"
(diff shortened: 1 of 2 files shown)


kangalio:
What prompted you to make this PR, by the way? The need to construct a `FrameworkContext`? If so, that should probably be a method on `Framework`. Or accessing `bot_id` from `Framework`? I think you'll only need to do that in the setup function, in which you can also use `ready.user.id`.

PR #130, opened 2022-11-30
[Feature] Allow to bypass command checks for owners

peanutbother:
This feature introduces a new option in FrameworkOptions called skip_checks_for_owners which allows owners to skip command dispatch checks to bypass permission checks.

 Added usage documentation in [framework_usage](https://github.com/peanutbother/poise/pull/examples/framework_usage/main.rs).

### DIFF
--- a/examples/framework_usage/main.rs
+++ b/examples/framework_usage/main.rs
@@ -134,6 +134,9 @@ async fn main() {
                 Ok(true)
             })
         }),
+        /// Enforce command checks even for owners (enforced by default)
+        /// Set to true to bypass checks, which is useful for testing
+        skip_checks_for_owners: false,
         event_handler: |_ctx, event, _framework, _data| {
             Box::pin(async move {
                 println!("Got an event in event handler: {:?}", event.name());
--- a/src/dispatch/common.rs
+++ b/src/dispatch/common.rs
@@ -77,6 +77,11 @@ async fn check_permissions_and_cooldown_single<'a, U, E>(
     ctx: crate::Context<'a, U, E>,
     cmd: &'a crate::Command<U, E>,
 ) -> Result<(), crate::FrameworkError<'a, U, E>> {
+    // Skip command checks if `FrameworkOptions::skip_checks_for_owners` is set to true
+    if ctx.framework().options.skip_checks_for_owners {
+        return Ok(());
+    }
+
     if cmd.owners_only && !ctx.framework().options().owners.contains(&ctx.author().id) {
         return Err(crate::FrameworkError::NotAnOwner { ctx });
     }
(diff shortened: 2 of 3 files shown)


examples/framework_usage/main.rs
```diff
@@ -134,6 +134,9 @@ async fn main() {
                 Ok(true)
             })
         }),
+        /// Enforce command checks even for owners (enforced by default)
+        /// Set to true to bypass checks, which is useful for testing
+        skip_checks_for_owners: false,
```
kangalio, line 139: By the way you don't need to explicitly write this field here because `false` is the default value anyways, so it be filled in by the `Default::default()` below

src/dispatch/common.rs
```diff
@@ -148,8 +148,12 @@ async fn check_permissions_and_cooldown_single<'a, U, E>(
...
     }
 
-    // Only continue if command checks returns true. First perform global checks, then command
-    // checks (if necessary)
+    // Only continue if command checks returns true or
+    // `FrameworkOptions::skip_checks_for_owners` is set to true
+    // First perform global checks, then command checks (if necessary)
+    if ctx.framework().options.skip_checks_for_owners {
+        return Ok(());
+    }
```
kangalio: Only custom checks and cooldowns are skipped, not guild, DM, NSFW, permission, and bot permission checks. Is this intended?

src/structs/framework_options.rs
```diff
@@ -23,6 +23,8 @@ pub struct FrameworkOptions<U, E> {
     /// If individual commands add their own check, both callbacks are run and must return true.
     #[derivative(Debug = "ignore")]
     pub command_check: Option<fn(crate::Context<'_, U, E>) -> BoxFuture<'_, Result<bool, E>>>,
+    /// If set to true, skips command checks if command was issued by [`FrameworkOptions::owners`]
```
kangalio, line 26: If it's intended that not all checks are skipped (see above), could you clarify this text to say which checks in particular are skipped?

examples/framework_usage/main.rs
peanutbother: just wanted to be verbose for documentation

src/dispatch/common.rs
peanutbother: actually forgot about that. where should I look for those checks?

kangalio: Look above 😄 It's all in that one function

src/dispatch/common.rs
```diff
@@ -77,6 +77,11 @@ async fn check_permissions_and_cooldown_single<'a, U, E>(
     ctx: crate::Context<'a, U, E>,
     cmd: &'a crate::Command<U, E>,
 ) -> Result<(), crate::FrameworkError<'a, U, E>> {
+    // Skip command checks if `FrameworkOptions::skip_checks_for_owners` is set to true
+    if ctx.framework().options.skip_checks_for_owners {
+        return Ok(());
+    }
```
kangalio, line 85: Wait.. this skips checks unconditionally for everyone if skip_checks_for_owners if enabled :p

peanutbother: duhh I'm not in my best form today, sorry for the noise

peanutbother: actually had that check in there before but my commit got lost. anyway got that fixed now

PR #133, opened 2022-12-15
Make builtins::servers aware of teams.

NotNorom:
I was calling the servers builtin and noticed it didn't recognize me as bot owner. I realized that's because the bot is part of a team...
For now this is just a "is author part of the team" check because discord does not different roles for team members.


kangalio:
Thank you for the PR, good catch! The pre-command owner check uses `ctx.framework().options().owners.contains`. Could you replace builtins::servers' check with that, for consistency, brevity, ability to add other owners, and to save an HTTP request? (Not sure why builtins::servers wasn't already using it)

NotNorom:
Done!

kangalio:
Thanks!

I should look into that weird MSRV failure sometimes

PR #136, opened 2023-01-01
Fix clippy warnings

GnomedDev:
Just thought I'd drive by and drop off some clippy fixes

### DIFF
--- a/examples/fluent_localization/translation.rs
+++ b/examples/fluent_localization/translation.rs
@@ -45,8 +45,8 @@ pub fn format(
 }
 
 /// Retrieves the appropriate language file depending on user locale and calls [`format`]
-pub fn get<'a>(
-    ctx: Context<'a>,
+pub fn get(
+    ctx: Context<'_>,
     id: &str,
     attr: Option<&str>,
     args: Option<&fluent::FluentArgs<'_>>,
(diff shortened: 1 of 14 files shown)


kangalio:
Thanks! Could you change back the `{foo}` format shorthand to `{}, foo`? I'm not a big fan of the inconsistency it causes because it's only applicable sometimes

GnomedDev:
Can you point to some inconsistencies, so I can add a couple bindings to clean them up? I much prefer following the standard formatting then introducing inconsistencies with the rest of the ecosystem.

kangalio:
> Can you point to some inconsistencies, so I can add a couple bindings to clean them up?

Not sure what you mean with bindings, but with inconsistencies I'm referring to the fact only identifier expressions can be used in the new shorthand, not other expressions. So instead of a single uniform syntax, there's now two choices and you have to transform between the two when changing the value to be printed. This is why I like to avoid the new shorthand

GnomedDev:
By bindings, I meant replacing 
```rust
format!("Hello {}!", user.name)
```
with
```rust
let user_name = user.name;
format!("Hello {user_name}")
```
so one style is always used.

kangalio:
I see. I'd really prefer if the one style to be used everywhere stayed the old one. Having to add dummy let bindings for simple expressions seems convoluted and annoying to do in my opinion

kangalio:
Closing due to inactivity and because clippy fixes are not that important

PR #138, opened 2023-01-01
Feature guard pagination example

GnomedDev:
This fixes compilation of poise with no default features, allowing CI to pass again.


kangalio:
Ngl such a feature guard shouldn't be needed in the first place but I don't feel like fixing the underlying issue rn

PR #143, opened 2023-02-23
Add missing commas in macros for localizations

max-m:
When you provide multiple localized names / descriptions the macros generate code like `vec![ ("foo", "bar") ("baz", "qux") ]`.
This PR adds the missing commas between the HashMap / Vec elements.


kangalio:
Thank you, nice catch!

PR #173, opened 2023-06-19
Added Support for Opening Modals as an MCI Response

G0ldenSp00n:
This lets users open modals as a response to a button press, instead of the current system that only allows them to be opened through a brand new context (ie. the immediate response of a slash command).

### Description
- Enabled opening modals on an MessageComponentInteraction instead of a directly
- Added Example Usage


kangalio:
I pushed some changes and included details and rationale in the commit description. Please tell me what you think about them @g0ldensp00n

examples/feature_showcase/modal.rs
```diff
@@ -15,3 +15,36 @@ pub async fn modal(ctx: poise::ApplicationContext<'_, Data, Error>) -> Result<()
...
+            .timeout(std::time::Duration::from_secs(120))
+            .filter(move |mci| mci.data.custom_id == "open_modal")
+            .await
+    {
+        let data =
+            poise::execute_modal_on_component_interaction::<MyModal>(ctx, mci, None, None).await?;
+        println!("Got data: {:?}", data);
+    }
+    Ok(())
+}
```
G0ldenSp00n, line 50: Yea, makes sense to me to have these live in the same file.  Again just didn't want to complicate things, was mostly trying to maintain the existing structures.

G0ldenSp00n (commented):
> - Remove Modal::execute_component_interaction function in favor of free-standing execute_modal_on_component_interaction function - I want to avoid an explosion of utility trait functions for any combination of defaults yes/no, component yes/no... That said, I realize we're already halfway in this mess with execute and execute_with_defaults both existing so I'd also be open to go all the way and add execute_component_interaction and execute_component_interaction_with_defaults

I fully agree with just letting the function live on it's own.  I only did it that way because all the other's were nested in the trait like that, and I didn't want to complicate the codebase more.  But I prefer it living on it's own.

Generally I agree with all the other changes you made.  I think it would also be good to add some example usage into the docs as well. Rather than everything living in the example folder like it does currently, just so you can see usage examples in your IDE hints but that can live in another PR.  Since it is needed everywhere right now, and I don't want this feature addition to be super bloated.

kangalio:
> I think it would also be good to add some example usage into the docs as well. Rather than everything living in the example folder like it does currently, just so you can see usage examples in your IDE hints

That would be nice. On the other hand, putting Discord bot commands into doc tests makes them annoying to test. I hope this trade-off is made obsolete by [rustdoc scraped examples](https://doc.rust-lang.org/rustdoc/scraped-examples.html)

G0ldenSp00n:
Anything more I need to do for this to be mergable?  I'm not super sure what the failing test is about, or if there is something I need to do in order to fix it?

PR #174, opened 2023-07-09
fix the behavior of poise::reply::builder::CreateReply.reply

B-2U:
according to the doc of the `reply` method [link](https://docs.rs/poise/latest/poise/reply/struct.CreateReply.html#method.reply)
when `reply(true)`, `allowed_mentions(|m| m.replied_user(boll)` should be `true` by default, but I noticed that I still have to manually enable it to get the ping
image
```rust
ctx.send(|builder| {
    builder
        .content(response)
        .reply(true)
        .allowed_mentions(|m| m.replied_user(true))
})
.await?;
ctx.send(|builder| {
    builder
        .content(response)
        .reply(true)
        // .allowed_mentions(|m| m.replied_user(true))
})
.await?;
```
so my approach is modify the `to_prefix` method in `builder.rs`
I also moved the condition check `if reply{..}` to above the `if let Some(allowed_mentions) = allowed_mentions {...}`, because compiler get a partial moved error that `allowed_mentions` will be moved in

please let me know if this is work as intended or there's some better way, thx!

### DIFF
--- a/src/reply/builder.rs
+++ b/src/reply/builder.rs
@@ -249,14 +249,6 @@ impl<'att> CreateReply<'att> {
             m.content(content);
         }
         m.set_embeds(embeds);
-
-        if reply {
-            m.reference_message(invocation_message);
-            if allowed_mentions.is_none() {
-                m.allowed_mentions(|m| m.replied_user(true));
-            };
-        }
-
         if let Some(allowed_mentions) = allowed_mentions {
             m.allowed_mentions(|m| {
                 *m = allowed_mentions;
@@ -269,6 +261,10 @@ impl<'att> CreateReply<'att> {
                 c
             });
         }
+        if reply {
+            m.reference_message(invocation_message);
+        }
+
         for attachment in attachments {
             m.add_file(attachment);
         }
--- a/src/reply/builder.rs
+++ b/src/reply/builder.rs
@@ -90,6 +90,9 @@ impl<'att> CreateReply<'att> {
     /// [`serenity::CreateAllowedMentions::replied_user`] set to false.
     pub fn reply(&mut self, reply: bool) -> &mut Self {
         self.reply = reply;
+        let mut allowed_mentions = serenity::CreateAllowedMentions::default();
+        allowed_mentions.replied_user(true);
+        self.allowed_mentions = Some(allowed_mentions);
         self
     }
 }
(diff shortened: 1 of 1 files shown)


kangalio:
Thank you for finding that issue and taking the time to create a PR 

As far as I can tell, this behavior is a result of an intended feature, but the docs of CreateReply.reply are wrong

Poise has a feature to set the allowed mentions globally for all messages sent by the bot (FrameworkOptions.allowed_mentions) which is set the "disable all pings except direct user pings" by default. As a result, reply pings are disabled by default too. So to enable reply pings you need to enable allowed mentions, either in CreateReply directly, or globally in FrameworkOptions

Please tell me what you think about this behavior - if it is unintuitive to library users, we can change it. If the behavior is ok, we can just update the docs

B-2U:
in my opinion it will be more intuitive if `FrameworkOptions.allowed_mentions` is true by default, just like when we are replying to other users in chat.
but I can understand that no ping is more like what slash commands did.

I'll say update the `reply` doc will be good, maybe also mention the existence of  `FrameworkOptions.allowed_mentions` inside.

kangalio:
> in my opinion it will be more intuitive if FrameworkOptions.allowed_mentions is true by default

Note that allowed_mentions is not a simple on/off setting - you can toggle mentions for
1. direct user pings
2. role pings
3. @ everyone pings
4. reply pings

I'd like 2 and 3 to stay switched off by default for safety. I assume with "allowed_mentions is true by default" you mean to set the reply pings part of allowed_mentions to true by default?

B-2U:
> you mean to set the reply pings part of allowed_mentions to true by default

ah you are right, I just realized that the `FrameworkOptions.allowed_mentions` also determine whether commands will be triggered by mentioning bot itself or roles etc. as prefix, and it's enabled by default. 

so I have a worry it will confuse library users when they try to enable `reply pings` in `FrameworkOptions`, but also disable the `f.empty_parse().parse(serenity::ParseValue::Users);` (which is the default)  unintentionally

B-2U:
> so I have a worry it will confuse library users when they try to enable reply pings in FrameworkOptions, but also disable the f.empty_parse().parse(serenity::ParseValue::Users); (which is the default) unintentionally

just make a quick test and seems it won't happen, sry about it

B-2U:
> ah you are right, I just realized that the FrameworkOptions.allowed_mentions also determine whether commands will be triggered by mentioning bot itself or roles etc. as prefix, and it's enabled by default.

And i totally misunderstood what `allowed_mentions` mean, 
anyway, yes, I mean the 4. reply pings

B-2U:
i guess that's all, thank you so much

kangalio:
> I just realized that the FrameworkOptions.allowed_mentions also determine whether commands will be triggered by mentioning bot itself or roles etc. as prefix, and it's enabled by default

No, that is not controlled by allowed_mentions; that is controlled by PrefixFrameworkOptions.mention_as_prefix

> just make a quick test and seems it won't happen, sry about it

Not sure what you even meant, but since it doesn't happen, I guess it's fine (:

PR #175, opened 2023-07-09
Add CooldownContext struct to make Cooldowns usable in more situations

scottbot95:
### Problem
I was wanting to implement cooldowns in some custom event handler, but unfortunately it is not practical to construct a `poise::Context` manually which is required to use the `Cooldowns` struct.

### Proposal
This PR refactors `Cooldowns` slightly to add an extra layer of abstraction around the `poise::Context` to make `Cooldowns` easier to use outside of a prefix/application command.

### DIFF
--- a/macros/src/command/prefix.rs
+++ b/macros/src/command/prefix.rs
@@ -78,7 +78,7 @@ pub fn generate_prefix_action(inv: &Invocation) -> Result<proc_macro2::TokenStre
             ))?;
 
             if !ctx.framework.options.manual_cooldowns {
-                ctx.command.cooldowns.lock().unwrap().start_cooldown(ctx.into());
+                ctx.command.cooldowns.lock().unwrap().start_cooldown(ctx.cooldown_context());
             }
 
             inner(ctx.into(), #( #param_names, )* )
--- a/macros/src/command/slash.rs
+++ b/macros/src/command/slash.rs
@@ -191,7 +191,7 @@ pub fn generate_slash_action(inv: &Invocation) -> Result<proc_macro2::TokenStrea
             })?;
 
             if !ctx.framework.options.manual_cooldowns {
-                ctx.command.cooldowns.lock().unwrap().start_cooldown(ctx.into());
+                ctx.command.cooldowns.lock().unwrap().start_cooldown(ctx.cooldown_context());
             }
 
             inner(ctx.into(), #( #param_identifiers, )*)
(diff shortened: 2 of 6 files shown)


kangalio (commented):
Thank you for this PR!

A few requests: Could you
- add a comment above CooldownContext explaining that it was factored out of Context to make cooldowns usable outside a poise::Context context
- target this PR to the `next` branch instead, since this is a breaking change
- remove the From impls to convert (Prefix, Application)Context to CooldownContext in favor of a struct method called `.cooldown_context()` (see https://github.com/serenity-rs/poise/blob/5d02b8757d30e4588c193c5ba06e38806bbc1021/src/structs/context.rs#L74)

scottbot95:
Thanks for the quick review! I can make those suggested changes. I was kinda skeptical if the `From` implementations were really the best way to make it easy to create.

Do you have opinions on implementing `From<serenity::Message>`? I added that since that is the use-case I was planning to use this for but I could see the argument that such a helper is out of scope of the library and should just be done by the user.

scottbot95:
Updates made. The `current` branch on my fork is now really just the `next` branch cuz I didn't want to re-create the PR and lose some context. But who cares; I'll just nuke my `current` branch after this PR if I decide I care enough lol.

kangalio:
Thank you for implementing my requests!

> Do you have opinions on implementing From<serenity::Message>? I added that since that is the use-case I was planning to use this for but I could see the argument that such a helper is out of scope of the library and should just be done by the user.

I agree with the thought process. I think it's ok to keep the impl here, because it doesn't bother in the docs and you can trivially reimplement it if you don't know it exists

scottbot95:
Hmm of course there were several linter errors. Linters seem to be the bane of my existence ;) Some of the errors don't seem related to my change but I went ahead and made cargofmt/clippy happy anyway.

scottbot95:
Hmmm all those failures passed when I tried to replicate the commands locally on my Windows machine or in the WSL2. And the errors don't seem related to my code at all. Not sure how best to proceed from here.

----

Tangential question if you know: What was the reasoning behind using an `OrderedMap` for storing the cooldowns instead of just a `std::HashMap`? It doesn't seem like we ever actually care about that order is preserved. The only other thing I could think of is that maybe it's a performance/memory thing under the assumption that these maps will normally be small so it's better to search a list than to use `HashMap`? If so is that really an assumption we want to make? Seems like that would just be deciding ahead of time that this library will never be used at scale.

kangalio:
You're right those failures don't seem related to this PR. Just plain code rot >:(

---

I checked and I actually can't figure out why I used OrderedMap for the cooldowns. I say we change it to HashMap and if nothing breaks, keep it

PR #180, opened 2023-07-10
Convert Cooldowns to use HashMap instead of OrderedMap for better scaling

scottbot95:
Follow-up from discussion in #175:

I have replaced the `OrderedMap` with `HashMap` in `Cooldowns` and everything seems to be fine (all tests pass and no issues in some simple tests with my WIP bot). 

I did have to remove `#[derive(Hash)]` from `Cooldowns` as `HashMap` does not implement `Hash`, however it seems weird that you would ever want the hash of a `Cooldowns` object since that would be highly volatile depending on the state of the cooldowns. If we think it's important to preserve the `Hash` impl on `Cooldowns`, I would suggest we a `BTreeMap` instead of the `OrderedMap` since we never iterate through the entire map and lookups/inserts will be much more performant on a `BTreeMap` at scale.

This also leaves `OrderedMap` with only one use-case which just uses `get_or_insert_with` and `into_iter()`. This meant I had to add `#[allow(dead_code)]` to `get` and `insert` of `OrderedMap` since those aren't used anymore. We could just delete the functions but I didn't want to do that since IMO it would be weird to have a `*Map` type with only a special insert and iterate. The other option would be to just remove `OrderedMap` entirely and use a `BTreeMap` in places where order matters and `HashMap`s elsewhere, but I don't see any issue with keeping the `OrderedMap` for the help use-case since order does matter and the map is probably small enough that it doesn't matter (probably even large cases will have <100 commands).
https://github.com/serenity-rs/poise/blob/6cc521a9b0398384c2be15ef4a5d8d1b9e52364a/src/builtins/help.rs#L293-L298 
https://github.com/serenity-rs/poise/blob/6cc521a9b0398384c2be15ef4a5d8d1b9e52364a/src/builtins/help.rs#L305

### DIFF
--- a/src/cooldown.rs
+++ b/src/cooldown.rs
@@ -2,7 +2,7 @@
 
 use crate::serenity_prelude as serenity;
 // I usually don't really do imports, but these are very convenient
-use crate::util::OrderedMap;
+use std::collections::HashMap;
 use std::time::{Duration, Instant};
 
 /// Subset of [`crate::Context`] so that [`Cooldowns`] can be used without requiring a full [Context](`crate::Context`)
@@ -38,7 +38,7 @@ pub struct CooldownConfig {
 ///
 /// You probably don't need to use this directly. `#[poise::command]` automatically generates a
 /// cooldown handler.
-#[derive(Default, Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Default, Clone, Debug, PartialEq, Eq)]
 pub struct Cooldowns {
     /// Stores the cooldown durations
     cooldown: CooldownConfig,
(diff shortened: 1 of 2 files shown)


kangalio:
I wasn't really aware of BTreeMap as an easy ordered map type - I think removing poise's OrderedMap implementation in favor of BTreeMap is a good idea

scottbot95:
Haha ok one more PR coming right up then!

scottbot95:
Oh actually BTreeMap is based off key order, not insertion order 
> Iterators obtained from functions such as [BTreeMap::iter](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.iter), [BTreeMap::values](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.values), or [BTreeMap::keys](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.keys) produce their items in order by key, and take worst-case logarithmic and amortized constant time per item returned.

So that would actually introduce a behavior change of sorting all command categories by the `Ord` impl of `String` (which presumably is alphabetically) instead of order of appearance in `FrameworkOptions.commands`. This is maybe fine? If you're ok with that I can still make the PR otherwise I think we should just leave it using the `OrderedMap`.

kangalio:
I'd prefer to keep the order of appearance in FrameworkOptions.commands. Also, isn't OrderedMap used in a couple more places where insertion order is needed?

scottbot95:
> I'd prefer to keep the order of appearance in FrameworkOptions.commands.

Yeah makes perfect sense. No more follow-up PRs then :(

> Also, isn't OrderedMap used in a couple more places where insertion order is needed?

Doesn't [look like it?](https://github.com/search?q=repo%3Aserenity-rs%2Fpoise%20OrderedMap&type=code). There is a TODO I suppose lol

kangalio:
> No more follow-up PRs then :(

You're free to tackle the open issues, or any TODOs in the code ;)

PR #181, opened 2023-07-15
Fix failing CI build

scottbot95:
- Update serenity 0.11.5 -> 0.11.6
- Update proc_macro2 1.0.47 -> 1.0.64
- Disambiguate several serenity re-exports
- Temporarily disable serenity/simdjson test since feature is broken upstream (serenity-rs/serenity#2474)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,10 +39,12 @@ jobs:
           - name: no features
             # don't test examples because they need collector feature
             command: cargo test --no-default-features --features serenity/rustls_backend --lib --tests
-          
-          - name: all features + simdjson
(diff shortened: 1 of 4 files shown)


scottbot95:
I think it's debate whether we disable the `simdjson` step. If we think that's fine I'll cut a ticket to track the follow-up to re-enable

kangalio:
Thank you!

You targeted the next branch - what are the breaking changes in this PR? Is it the import disambiguations?

scottbot95:
Yes the import disambiguations are potentially breaking if people were using the imports through `poise` before the ambiguity was introduced; although I tried to pick sensible disambiguations so that hopefully shouldn't be the case. But the real reason I targeted `next` was I simply didn't realize there were similar errors in `current` lol (although not as many it would seem?)

PR #183, opened 2023-07-16
add `reply()` for `poise::context`

B-2U:
the only problem is the name, `reply_reply` is definitly confusing, but I don't have a better idea than it :(

### DIFF
--- a/src/reply/send_reply.rs
+++ b/src/reply/send_reply.rs
@@ -5,7 +5,7 @@ use crate::serenity_prelude as serenity;
 /// Send a message in the given context: normal message if prefix command, interaction response
 /// if application command.
 ///
-/// If you just want to send a string, use [`say_reply`].
+/// If you just want to send a string, use [`say_reply`] or [`reply_reply`].
 ///
 /// Note: panics when called in an autocomplete context!
 ///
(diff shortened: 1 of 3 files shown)


src/reply/send_reply.rs
```diff
@@ -44,6 +44,18 @@ pub async fn say_reply<U, E>(
     send_reply(ctx, |m| m.content(text.into())).await
 }
 
+/// Shorthand of [`send_reply`] for text-only messages, and reply to the reference message for [`crate::PrefixContext`]
+///
+/// for [`crate::ApplicationContext`], this is the same as [`say_reply`].
+///
+/// Note: panics when called in an autocomplete context!
+pub async fn reply_reply<U, E>(
```
kangalio: I say remove `reply_reply` - those plain functions for replying have become kinda useless anyways (`say_reply(ctx, "...")` is inferior to `ctx.say("...")`) so it's not really important to add this new utility feature to them

src/structs/context.rs
```diff
@@ -131,6 +131,21 @@ context_methods! {
...
     }
 
+    /// Shorthand of [`crate::reply_reply`]
+    ///
+    /// reply to the reference message in prefix commands
+    ///
+    /// for slash commands, this is the same as [`say`].
+    ///
+    /// Note: panics when called in an autocomplete context!
+    await (say self text)
```
kangalio: Please change the method name here from say to reply, or the method implementations on PrefixContext and ApplicationContext will have the wrong name

kangalio (changes requested):

src/reply/send_reply.rs
B-2U: do you mean I should directly use `send_reply()` builder in `ctx.reply()` ?

kangalio: Yes, or even `ctx.send(|b| b.content(text).reply(true))`

B-2U:
Hmm it should be good now, do i need to click something after mark the requests as resolved?

kangalio (approved):

PR #185, opened 2023-07-16
Allow dynamic `CooldownConfig`

scottbot95:
Introduce `CooldownConfigProvider` to allow dynamic `CooldownConfig`.

Shouldn't be particularly breaking beyond #175 since `CooldownConfig` implements `CooldownConfigProvider`

### DIFF
--- a/examples/feature_showcase/checks.rs
+++ b/examples/feature_showcase/checks.rs
@@ -8,7 +8,8 @@ Poise supports several pre-command checks (sorted by order of execution):
 - cooldowns
 */
 use crate::{Context, Error};
-use poise::serenity_prelude as serenity;
+use poise::{serenity_prelude as serenity, CooldownConfig};
+use std::time::Duration;
 
 #[poise::command(prefix_command, owners_only, hide_in_help)]
 pub async fn shutdown(ctx: Context<'_>) -> Result<(), Error> {
@@ -86,7 +87,6 @@ pub async fn ferrisparty(ctx: Context<'_>) -> Result<(), Error> {
     Ok(())
 }
 
-/// Add two numbers
 #[poise::command(
     prefix_command,
     track_edits,
@@ -99,6 +99,29 @@ pub async fn ferrisparty(ctx: Context<'_>) -> Result<(), Error> {
     member_cooldown = 3,
 )]
 pub async fn cooldowns(ctx: Context<'_>) -> Result<(), Error> {
+    ctx.say("You successfully called the command").await?;
+    Ok(())
+}
+
+async fn random_cooldowns(ctx: Context<'_>) -> Result<bool, Error> {
+    ctx.command()
+        .cooldowns
+        .lock()
+        .unwrap()
+        .set_config(CooldownConfig {
+            global: Some(Duration::from_secs_f32(rand::random()) * 10),
+            user: Some(Duration::from_secs_f32(rand::random()) * 10),
+            guild: Some(Duration::from_secs_f32(rand::random()) * 10),
+            channel: Some(Duration::from_secs_f32(rand::random()) * 10),
+            member: Some(Duration::from_secs_f32(rand::random()) * 10),
+            __non_exhaustive: (),
+        });
+
+    Ok(true)
+}
+
+#[poise::command(prefix_command, track_edits, slash_command, check = "random_cooldowns")]
+pub async fn dynamic_cooldowns(ctx: Context<'_>) -> Result<(), Error> {
     ctx.say("You successfully called the command").await?;
     Ok(())
 }
--- a/src/cooldown.rs
+++ b/src/cooldown.rs
@@ -120,6 +120,11 @@ impl Cooldowns {
             self.member_invocations.insert((ctx.user_id, guild_id), now);
         }
     }
+
+    /// Updates the [`CooldownConfig`] in use by this [`Cooldowns`]
+    pub fn set_config(&mut self, config: CooldownConfig) {
+        self.cooldown = config;
+    }
 }
 
 impl<'a> From<&'a serenity::Message> for CooldownContext {


src/cooldown.rs
```diff
@@ -3,18 +3,22 @@
...
 use std::collections::HashMap;
+use std::fmt::Debug;
+use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 /// Subset of [`crate::Context`] so that [`Cooldowns`] can be used without requiring a full [Context](`crate::Context`)
 /// (ie from within an `event_handler`)
-#[derive(Default, Clone, PartialEq, Eq, Debug, Hash)]
-pub struct CooldownContext {
+#[derive(Clone, PartialEq, Eq, Debug, Hash)]
```
scottbot95: Remove `Default` here since we can't enforce `U: Default`. Probably doesn't make sense for this to be `Default` anyway.

src/cooldown.rs
```diff
@@ -34,14 +38,32 @@ pub struct CooldownConfig {
     pub __non_exhaustive: (),
 }
 
+/// Trait to allow loading custom [`CooldownConfig`]s based on the current [`CooldownContext`]
+#[async_trait::async_trait]
+pub trait CooldownConfigProvider<U, E>: Debug + Send + Sync {
```
scottbot95: Would be nice if we didn't need a trait here and could just use something like `Fn(&CooldownContext) -> BoxFuture<Result<CooldownConfig, E>>`. I tried that but couldn't get around needing to add a lifetime parameter to `Cooldowns` which forces on being added to `Command` and that seemed to impactful of a change to make.

src/cooldown.rs
```diff
@@ -3,18 +3,22 @@
...
+use std::fmt::Debug;
+use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 /// Subset of [`crate::Context`] so that [`Cooldowns`] can be used without requiring a full [Context](`crate::Context`)
 /// (ie from within an `event_handler`)
-#[derive(Default, Clone, PartialEq, Eq, Debug, Hash)]
-pub struct CooldownContext {
+#[derive(Clone, PartialEq, Eq, Debug, Hash)]
+pub struct CooldownContext<'a, U> {
```
scottbot95: Is there anything else we should add to this struct? This struct was factored out of `poise::Context` to be the minimal set of data needed to calculate cooldowns so it should be kept small and probably shouldn't have `non_exhaustive` to enable its use in the widest range of use-cases.

Potential things available within an `event_handler that we could add:
| Data| Comments|
|--------|--------|
| The `Event` itself | Probably not directly relevant to Cooldowns |
| The `FrameworkContext`| Only relevant fields are the user data and maybe the bot_id |
| The `serentiy::Context` | Probably not relevant for computing the cooldowns |

kangalio:
Skimming over the PR, I'm a bit weary of the amount of complexity added for a pretty small feature

Have you considered implementing these dynamic cooldowns in your bot with custom code running inside your bot command function, rather than framework level support?

If that's not possible for some reason, I'll look into why exactly the Fun(...) -> BoxFuture approach isn't working, and maybe even use a different approach entirely (just wrapping CooldownConfig in Mutex; dynamic cooldowns would then be a matter of a pre command hook)

scottbot95:
IMO it's not _that_ much complexity since a lot of it is just boilerplate stuff, but I agree it is a somewhat controversial change. I am definitely open to other ideas.

I certainly could implement such a feature within my own bot code, but I figured I'd either do something like manage multiple `Cooldowns` for the various configs (not ideal since the timers can be re-used) or mostly re-implementing the `Cooldowns` type within my own project so I might as well make this PR to perhaps contribute upstream.

It may just be a limitation of my Rust knowledge, but it seems like using `Fn` instead requires `Cooldowns` to take on a lifetime parameter `'a` which would then `'a` to be propagated up to `Command`. I did not fully investigate the feasibility of adding a lifetime parameter to `Command`, but that will only add to the complexity which is why I shied away from that route at first.

src/cooldown.rs
kangalio: Did you try `for<'a> Fn(&'a CooldownContext) -> BoxFuture<'a, Result<CooldownConfig, E>>`?

scottbot95: I did see some similar uses elsewhere in the code. I was hitting some issues with that, but that's likely just because I don't *really* understand what `for<'a>` does (it means the trait bound needs to work for all lifetimes `'a` but I haven't internalized what the implications of that are).

I took another crack at it and totally got it (mostly) working. Although functions passed in through `#[poise::command(cooldown_provider)]` can't be `async` and need to manually handle the `BoxFuture` stuff. Still messing around to see if I can get that working nicely.

kangalio:
I'm currently on vacation, I'll get back to this in a few weeks

scottbot95:
Refactored to use a `Fn` instead of a custom trait for the provider. For reasons beyond my current understanding it requires Rust 1.67+ even though there's nothing in the 1.67.0 changelog that seems like it would be related. Bumping the MSRV to 1.67 is probably not something we really want to do so this push is more to share the code in case anyone has ideas on how to fix it so it works with 1.60. If we can't resolve it, then I'll just close this PR and end up with something pretty custom in my own bot.

Edit: the MSRV test pass here because we don't run them with the `--examples` flag

kangalio:
> **kangalio:** maybe even use a different approach entirely (just wrapping CooldownConfig in Mutex; dynamic cooldowns would then be a matter of a pre command hook)

I've tried this locally by adding literally just a method `set_config(&mut self, CooldownConfig)` to `Cooldowns` and it compiles. You would use it via `Command.cooldowns.lock().set_config(...)`, where the `Command` instance can be accessed in a pre-command hook or whereever. What do you think of this?

scottbot95:
That certainly seems like it could work. I suppose it's really not much different of an interface too since it would more-or-less just be swapping the `cooldown_provider` attribute for an extra `check` hook. Seems like a much less intrusive change too. It's not quite as clean a user experience as I was hoping, but it probably worth it for the significantly reduced complexity.

examples/feature_showcase/checks.rs
```diff
@@ -103,6 +103,29 @@ pub async fn cooldowns(ctx: Context<'_>) -> Result<(), Error> {
...
+        .cooldowns
+        .lock()
+        .unwrap()
+        .set_config(CooldownConfig {
+            global: Some(Duration::from_secs_f32(rand::random()) * 10),
+            user: Some(Duration::from_secs_f32(rand::random()) * 10),
+            guild: Some(Duration::from_secs_f32(rand::random()) * 10),
+            channel: Some(Duration::from_secs_f32(rand::random()) * 10),
+            member: Some(Duration::from_secs_f32(rand::random()) * 10),
+            __non_exhaustive: (),
```
scottbot95, line 117: Kinda awkward to force users to include this. Should we perhaps make this struct not `non_exhaustive`? Or create a `new` function?

kangalio: Good point. Should probably add `Command.config(&self) -> CooldownConfig` too so you can modify the config incrementally instead of having to build one from scratch

I just checked and CooldownConfig has a Default impl too, so that's also a possibility to avoid the __non_exhaustive

kangalio:
Can you explain how you'll use dynamic cooldowns in your bot? I'm struggling to see cooldowns dynamically changing per invocation as anything more than a "why though - because we can" gimmick right now and it would help me to have a fuller picture of this feature.

Possibly the random_cooldowns example can be adapted to a more realistic example, to explain the usefulness of dynamic cooldowns to other users too

scottbot95:
My bot, [FaultyBot](https://github.com/scottbot95/faultybot), allows users with sufficient permission to manage guild/channel/member settings. One of the settings will be the cooldowns for various actions. 

I agree a more realistic example would be better. Unfortunately I also can't really come up with a use-case other than my dynamic settings thing. And there's lot of extra infra we would need to have a meaningful example for settings

kangalio:
> My bot, [FaultyBot](https://github.com/scottbot95/faultybot), allows users with sufficient permission to manage guild/channel/member settings. One of the settings will be the cooldowns for various actions.

Ohh I didn't think of that. Now this feature doesn't look arbitrary to me anymore

kangalio:
Btw I'd have appreciated if you didn't always force-push ;) (I'm struggling to check how you originally implemented the PR, for inspiration purposes)

scottbot95:
> Ohh I didn't think of that. Now this feature doesn't look arbitrary to me anymore

Yeah I thought it would be a nice feature to have. I'm trying to make this bot actually meaningful usable at scale (servers with 10K+ users). Related, I'm having to come up with my own permission system too since I need custom permissions (like the permission to change settings in a guild). If you feel like that's something that has any place at all in poise and I come up with an implementation that isn't just horrendously hacky, I'd be down to submit a PR for that as well.

> Btw I'd have appreciated if you didn't always force-push ;) (I'm struggling to check how you originally implemented the PR, for inspiration purposes)

Ah sorry about that lol. It's a habit I've picked up from my job cuz we like to keep the history as clean as possible. We have a custom review tool that lets us look at previous revisions though. What would you prefer in a case like this were the original implementation probably shouldn't end up in history? Just commit over it and we'll squash when we merge the PR or something? I'm relatively new to this whole OSS contributions thing lol

I actually saved a local branch with the old impl. Just pushed it in case you haven't managed to find the old commit already https://github.com/scottbot95/poise/tree/cd-config-provider-old

kangalio:
I implemented a slightly different design and pushed it to the `current` branch, along with an example in the `examples/` directory c9c037397afe5fabd56a3adff2ac3a59e52b68b4. Feel free to tell me if that works for your use case

scottbot95:
Yeah that would certainly work. Sadly, my bot's use-case requires the new `CooldownContext` from the `next` branch since it will be using the `CooldownTracker` from an event_handler so it doesn't have a full `poise::Context`. I will just close this PR.

I see in `current` there are comments about migrating to the "v2" APIs which require the `CooldownConfig` to be passed in. What is the plan for that for `Command`s? Just make the `Command` itself store the `CooldownConfig`?

kangalio:
> What is the plan for that for Commands? Just make the Command itself store the CooldownConfig?

Yeah, probably. I should have migrated the project internals to the new APIs already tbh

PR #186, opened 2023-07-19
Two dependency bumps

OverzealousLotus:
- Bump `tokio` to `1.25.1`. Running `cargo audit` reveals a denied vulnerability. I completely forgot to copy the rustsec link for this one. 
- Bump `env_logger` to `1.10.0`. Running `cargo audit` reveals a warned against vulnerability. See https://rustsec.org/advisories/RUSTSEC-2021-0145

Apologies for the unverified commit. I'm having issues with VSCode. :sweat:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,17 +54,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


kangalio:
Not sure how much this matters, since nothing stops users from updating the version of tokio on their own (simple `cargo update` will update transitive dependencies too) and `env_logger` is only used in examples anyways. But I suppose it can't hurt either (except for if someone requires an old version of tokio for some reason...? Which poise would refuse to support with this PR, even though it theoretically could)

PR #190, opened 2023-08-10
Merge changes from `current`

scottbot95:
- Merge in the changes from `current` branch.
- Remove deprecated methods from `CooldownTracker` (I did this as a sperate commit so it would be really easy to make it a separate PR if you would prefer that)
- Updated `Command` to new `CooldownTracker` methods. I included this on the `next` branch since it seems like it's somewhat of a breaking change since it's adding a field to `Command` and people _shouldn't_ be making those on their own but they could be


scottbot95:
Hmmm the error seems to be caused by f9fce51f but the error was masked in `current` due to the unrelated failures

scottbot95:
Oops pushed more changes prematurely to my forked `next` branch.

Bumped MSRV to 1.63 to allow build to succeed. Although the doc build is failing now for reasons? The `if: github.ref == 'refs/heads/current' || github.ref == 'refs/heads/next'` check seems to be getting confused by the fact that my fork's branch name is also `next`. But that seems like [a spec violation by github](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) so... 🤷 

Edit: Oh nvm I'm just confusing myself but looking at builds that happened on my fork. I didn't realize Github ran actions on forks. Perhaps I turned that on at some point?

PR #199, opened 2023-09-03
add simple example for event handler

### LINKED ISSUE #188: Request: Example of event_handler
Hi!

Looking at the Discord, almost every week someone asks about an explanation on how the event_handler works.

Would it be a good idea to just make an example using the event_handler where it is used to collect messages, message updates and so on?

gkuznik:
This counts the number of sent "pings" and replies with that number.
It's very simple, but I think this is enough to show what can be done with an event handler.

closes #188

### DIFF
--- a/examples/event_handler/main.rs
+++ b/examples/event_handler/main.rs
@@ -10,7 +10,7 @@ type Context<'a> = poise::Context<'a, Data, Error>;
 
 // Custom user data passed to all command functions
 pub struct Data {
-    pongs: AtomicU32,
+    poise_mentions: AtomicU32,
 }
 
 #[tokio::main]
@@ -32,7 +32,7 @@ async fn main() {
         .setup(move |_ctx, _ready, _framework| {
             Box::pin(async move {
                 Ok(Data {
-                    pongs: AtomicU32::new(0),
+                    poise_mentions: AtomicU32::new(0),
                 })
             })
         })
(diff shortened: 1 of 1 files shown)


examples/event_handler/main.rs
```diff
@@ -0,0 +1,70 @@
...
+    event: &Event<'_>,
+    _framework: poise::FrameworkContext<'_, Data, Error>,
+    data: &Data,
+) -> Result<(), Error> {
+    match event {
+        Event::Ready { data_about_bot } => {
+            println!("Logged in as {}", data_about_bot.user.name);
+        }
+        Event::Message { new_message } => {
+            if new_message.content == "!ping" {
```
kangalio: This is a bit of a misleading example because poise has built-in support for text commands, and if someone wanted to implement them manually they should use base serenity instead of poise.

Could you change the code to count, e.g. the number of.. all-uppercase messages, or messages containing a specific keyword, or whatever? I think that'd be a more realistic use case of the event handler

kangalio:
Thank you for tackling the missing example!

kangalio (changes requested):
Thank you for tackling the missing example! I have one nit

examples/event_handler/main.rs
gkuznik: Sure, changed it to count Poise.

kangalio: Haha some subtle self advertisement

kangalio:
As far as I can tell the CI errors are unrelated, so I think this PR can be merged. Or are you planning to force push some more changes @Friendly-Banana ?

gkuznik:
@kangalio Nope, I'm finished. A little bit of free advertising is always nice.

gkuznik:
@kangalio from my side this PR is ready. Can we merge :smile: ?

arqunis (approved):

PR #200, opened 2023-09-06
Fix modal acknowledgement

arqunis:
Discord has ceased accepting the `DeferredUpdateMessage` response type for acknowledging a modal, which has resulted in errors propping up when using modals via poise.

This fixes said errors by switching to the alternative `DeferredChannelMessageWithSource` type. In addition, the response to the modal is made ephemeral to prevent a "<Bot> is thinking..." message from appearing, waiting to be replaced by proper message content, which will never happen.

### DIFF
--- a/src/modal.rs
+++ b/src/modal.rs
@@ -65,7 +65,7 @@ async fn execute_modal_generic<
     // Send acknowledgement so that the pop-up is closed
     response
         .create_interaction_response(ctx, |b| {
-            b.kind(serenity::InteractionResponseType::DeferredUpdateMessage)
+            b.kind(serenity::InteractionResponseType::DeferredChannelMessageWithSource)
         })
         .await?;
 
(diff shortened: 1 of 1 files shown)


kangalio:
This will show a "This interaction failed" after 5 seconds in an ephemeral box for the invoking user, right?

arqunis:
> This will show a "This interaction failed" after 5 seconds in an ephemeral box for the invoking user, right?

That message is displayed when a client to Discord's API fails to respond within the 3 second timeframe upon receiving an interaction. With `DeferredChannelMessageWithSource`, for the user a "<bot> is thinking..." message is shown, which will last a few minutes before expiring.

kangalio:
Hm okay. Is there no way anymore to just acknowledge a modal submit without replying with a text?

arqunis:
> Hm okay. Is there no way anymore to just acknowledge a modal submit without replying with a text?

No, unless they would add a new response type just for acknowledging a modal.

kangalio:
That's sad.

The modal API in poise was built with the assumption that you can acknowledge a model submit without any further input needed. I suppose this workaround with the ephemeral loading indicator will have to do for now as a workaround

PR #201, opened 2023-09-12
paginate: use saved ID instead of regenerating each time

seqre:
`poise::builtins::paginate` uses `ctx.id()` to generate one common prefix for unique button IDs. However, instead of using that prefix, it generates a new ID for each button. This PR fixes this and uses the prefix ID generated once at the beginning.


kangalio:
Good catch

kangalio:
Actually, `ctx.id()` always gives the same result so this wasn't a bug fix but purely a source code aesthetics change. Still good tho

Issue #203, opened 2023-09-17
`poise::builtins::servers` sends error when message is longer that 2000 characters.


NotNorom:
Hi!

I am using `poise::builtins::servers` to look at statistics. It's really neat that you added this function :)
Recently I've run into the problem that the message that would be sent is too large resulting in this being the content of the message after running though the on_error function:
`Invalid Form Body (data.content: Must be 2000 or fewer in length.)`

What would a good approach be here?
Obviously I could just run my own implementation splitting the message, but I wanted to ask if it makes more sense to have those changes in poise.

Thanks!

kangalio:
Hi, thanks for the report. I think the best approach would be to have the function keep track of the message length itself and stop adding lines when 2000 chars is reached.

I never tackled this issue before because it didn't bother me so much and the function is just for convenience

PRs welcome as always ^^

PR #204, opened 2023-09-18
Respect 2000 character limit in builtins::servers

NotNorom:
In relation to #203 

I am _not_ very happy with how I have written this but uhhhh yeah. Feedback is welcome.


NotNorom:
LGTM

PR #210, opened 2023-10-20
fix compiling with latest serenity

jamesbt365:
image


jamesbt365:
nevermind something was messed up my end and I don't know how or why

PR #211, opened 2023-10-22
Add ignore_thread_creation to PrefixFrameworkOptions

seraxis:
When a new thread is created the title can be parsed as if it were a prefix command invocation from a regular message, this adds an option to `PrefixFrameworkOptions` to ignore these accidental command invocations by default while preserving old behaviour if desired.


kangalio:
Immaculate PR

PR #216, opened 2023-11-02
Prevent underscore binding lint

mattfbacon:
This lint is triggered when we have an unused parameter in the handler function. The code that's generated copies the identifier name from the function definition, but does use the identifier, so we need to disable this lint. That said, a better solution might be to add a fixed prefix like `poise_arg_` to each parameter name.


kangalio:
I wish Clippy and co would just ignore macro-generated code :pensive:

mattfbacon:
I would argue that macro-generated code should be held to the same standards as normal code. That said, disabling the lint is way easier than adding a prefix.

PR #224, opened 2023-11-24
Upgrade poise current to include poise next and serenity-next

GnomedDev:
Discussed on discord.


kangalio (commented):
Would have been easier to review with the "merge next into current" and "port to serenity 0.12" split into two PRs, but it's ok

src/modal.rs
```diff
@@ -27,7 +25,7 @@ pub fn find_modal_text(
 
         if text.custom_id == custom_id {
             let value = std::mem::take(&mut text.value);
-            return if value.is_empty() { None } else { Some(value) };
+            return value.and_then(|val| (!val.is_empty()).then_some(val));
```
kangalio: :raised_eyebrow: is this really better?

src/reply/send_reply.rs
```diff
@@ -138,19 +118,7 @@ async fn _send_prefix_reply<'a, U, E>(
...
-                // (*f = Default::default() won't do)
-                f.content("");
-                f.set_embeds(Vec::new());
-                f.components(|b| b);
-                f.0.insert("attachments", serenity::json::json! { [] });
-
-                reply.to_prefix_edit(f);
-                f
-            })
+            .edit(ctx.serenity_context, builder.to_prefix_edit())
```
kangalio: I see you sneakily moved the reset code from here into to_prefix_edit. This subtly changes the behavior of to_prefix_edit-serialized CreateReply instances: fields that are not set will not be left alone, but reset to default. I suppose that's fine, because semantically we're treating CreateReply as the specification of a stand-alone message, not an amend to a potentially-existent message.

src/structs/framework_error.rs
```diff
@@ -178,39 +192,14 @@ pub enum FrameworkError<'a, U, E> {
...
-            Self::DmOnly { ctx, .. } => ctx.serenity_context(),
-            Self::NsfwOnly { ctx, .. } => ctx.serenity_context(),
-            Self::CommandCheckFailed { ctx, .. } => ctx.serenity_context(),
-            Self::DynamicPrefix { ctx, .. } => ctx.serenity_context,
-            Self::UnknownCommand { ctx, .. } => ctx,
-            Self::UnknownInteraction { ctx, .. } => ctx,
-            Self::__NonExhaustive(unreachable) => match unreachable {},
-        }
-    }
-
```
kangalio: Why was this method removed?

src/modal.rs
GnomedDev: No, I don't know what I was thinking lol, lemme revert that

src/reply/send_reply.rs
GnomedDev: I moved it as to_prefix_edit is now creating the builder, so there was no other choice.

src/structs/framework_error.rs
GnomedDev: We would have to match over the entire serenity FullEvent to get the context out of it, as is.

src/reply/send_reply.rs
kangalio: Huh, no. `to_prefix_edit` could still leave fields unset (i.e. `None`), which would leave those aspects of the message unchanged when sending the builder to Discord.

src/structs/framework_error.rs
kangalio: Hmmm I see. Outright removing the method is really bad though for people who need it.

I think we should keep the method, but have it return `Option<&serenity::Context>`. It needs to be None one way or another because there may not be a serenity::Context available (FullEvent::Ratelimit). For now, it can just blanket return None if it would need to match over FullEvent. Later, when serenity gains a `FullEvent::context` method or similar, we can upgrade the method to work properly for those events that have an associated Context

I really don't want to remove the method outright without a designated replacement. You know, I'm thinking of the bot dev who couldn't care less about our API design and just wants his bot's features to keep working, and I don't want to pull the rug under those people by removing methods like these

GnomedDev: Is there not already the stopgap solution of calling `FrameworkError::ctx` then `Context::serenity_context`?

GnomedDev: Fixed by bumping to proper serenity 0.12, as the fullevent PR allowed this to exist again.

PR #234, opened 2023-12-30
Add missing non_exhaustive enum variant constructors to SlashArgError

webmsgr:


kangalio:
Actually, looking at this code - there _was_ a constructor function but it was `#[doc(hidden)]`, because I didn't intend it to be used outside the lib. For example the `Parse` variant is intended for poise's FromStr adapter (i.e. if a FromStr type is used as a parameter and it fails, `Parse` is thrown).

Why do you need to construct these framework-internal (intended as such, at least) variants in user code?

kangalio:
By the way, thank you! I sometimes forget to appreciate the effort of people submitting PRs

webmsgr:
I was using a command argument with a `FromStr` impl, but because the error type didn't implement Error (anyhow), I got some very confusing errors.  I tried to manually implement both `PopArgument` and `SlashArgument` (which some of the errors suggested) and ran into the issue I couldn't create a `SlashArgError::Parse` for when the `FromStr` fails inside the `SlashArgument` impl. I've basically scrapped the whole thing and just call `FromStr` manually now, so feel free to close if you want.

Two of the errors pictured below:
image
image

kangalio:
The FromStr error type was anyhow::Error? That's really sus. What crate is that, so I can raise an issue?

webmsgr:
Its part of the bots source code. Its probably not the best idea, but i was already using it and didn't want to over complicate things.

PR #240, opened 2024-01-17
Use serenity data

GnomedDev:
Blocked on #239. Moves poise to using serenity's client field, with an abort on Framework::init if the types do not match.


GnomedDev:
Merged manually.

PR #242, opened 2024-01-30
Take prefix as Cow<'static, str>

GnomedDev:

### DIFF
--- a/src/builtins/help.rs
+++ b/src/builtins/help.rs
@@ -1,7 +1,7 @@
 //! Contains the built-in help command and surrounding infrastructure
 
 use crate::{serenity_prelude as serenity, CreateReply};
-use std::fmt::Write as _;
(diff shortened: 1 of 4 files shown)


kangalio:
This is admittedly a tiny amount of churn, but for an even more minuscule amount of resource saving...

GnomedDev:
Small incremental changes are how we improve. This has bugged me for absolutely ages though

kangalio:
Alright. I personally feel that the type used for the prefix is so negligible for performance (`prefix` is static, and `dynamic_prefix` is only convenience for the purer `stripped_dynamic_prefix`), that the decision should be purely based on usability. And Rust's default go-to string type for everything that's not specialized is just String, hence to me String does feel cleaner than Cow<'static, str> here. But I don't really feel like bickering about this because I suppose in the end it's a small change

PR #251, opened 2024-03-17
Fix bug causing bot to reply to its own messages in event_handler example

Spacerulerwill:
In the event_handler example, sending message containing the word "poise" would cause the bot to reply with the message "Poise has been mentioned {n} times" like expected. However, since the bot's message itself contained the word "poise", it would continually reply to itself. 

I fixed it by adding a simple check to test the message_author's id against the bot's user id.


kangalio:
Thank you! The CI is still failing, but then we can merge

Spacerulerwill:
I have formatted the file properly with rustfmt and have ammended my commit

PR #255, opened 2024-03-26
.unwrap_or_default() instead of falling back to "<prefix>" in single-command help

### LINKED ISSUE #254: single-command help shows `<prefix>` for bot with no prefix
I don't want my bot to have a prefix; I want to invoke commands solely by pinging it.

Annoyingly, poise's built in help function shows the prefix as `<prefix>`

image

it would be acceptable if it was just nothing or `@bot_name ` (well, preferably configurable, but i'll take either)

It happens at [help.rs line 158](https://github.com/serenity-rs/poise/blob/current/src/builtins/help.rs#L158)

general help doesn't have this problem:
image

SIGSTACKFAULT:
fixes #254

### DIFF
--- a/src/builtins/help.rs
+++ b/src/builtins/help.rs
@@ -150,7 +150,7 @@ async fn help_single_command<U, E>(
             subprefix = Some(format!("  /{}", command.name));
         }
         if command.prefix_action.is_some() {
-            let prefix = match get_prefix_from_options(ctx).await.unwrap_or_default();
+            let prefix = get_prefix_from_options(ctx).await.unwrap_or_default();
             invocations.push(format!("`{}{}`", prefix, command.name));
             if subprefix.is_none() {
                 subprefix = Some(format!("  {}{}", prefix, command.name));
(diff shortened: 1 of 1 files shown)


kangalio:
Hi, thanks for the fix! Reading the code comments in the removed code, I would prefer you make the fallback string configurable, bc for dynamic prefixes, an empty string is not a good fallback either

SIGSTACKFAULT:
@kangalio `help_all_commands` already falls back to not showing any prefix [here](https://github.com/serenity-rs/poise/blob/current/src/builtins/help.rs#L272) so i copied that logic.

I think this whole file would need a bit of a rewrite

jamesbt365:
Just a heads up, help commands on `next` have been removed as the code quality is shockingly bad, and they aren't worth maintaining in the crate. #323 should explain it, and a potential path for moving forward.

GnomedDev:
Yeah, let's avoid adding more to this when it needs a rewrite.

PR #267, opened 2024-05-15
feat(macro): add group macro

TitaniumBrain:
Adds the poise::group macro, which lets you group commands in a struct.

Use `MyStruct::commands()` to get a vec of the commands.

See the doc of poise::group for more details.

Needs more documentation and examples.


macros/src/lib.rs
```diff
@@ -7,7 +7,10 @@ mod command;
 mod modal;
 mod util;
 
+use std::vec;
```
GnomedDev: Isn't this in the prelude?

macros/src/lib.rs
```diff
@@ -277,3 +280,140 @@ pub fn modal(input: TokenStream) -> TokenStream {
...
+    /// This is another command
+    #[poise::command(user_cooldown=2000)]
+    async fn command_two(ctx: Context<'_>) -> Result<(), Error> {
+        // code
+    }
+}
+```
+*/
+#[proc_macro_attribute]
+pub fn group(args: TokenStream, input_item: TokenStream) -> TokenStream {
```
GnomedDev, line 317: I would recommend splitting this out into a function like `group_impl` which returns a Result, so you can use `?` instead of this match spam. I implemented a similar thing in [bool_to_bitflags](https://github.com/GnomedDev/bool-to-bitflags/blob/master/src/error.rs) for inspiration.

macros/src/lib.rs
```diff
@@ -277,3 +280,140 @@ pub fn modal(input: TokenStream) -> TokenStream {
...
+                let cmd: proc_macro2::TokenStream = match command::command(new_args, function) {
+                    Ok(x) => x.into(),
+                    Err(e) => e.write_errors().into(),
+                };
+
+                impl_body = quote!(
+                    #impl_body
+                    #cmd
+                );
+                // repeated because if let chains are unstable <https://github.com/rust-lang/rust/issues/53667>
```
GnomedDev: Can't this be solved by returning `cmd` and `item` from the `if let` expression then doing the `quote!` at the end?

macros/src/lib.rs
```diff
@@ -277,3 +280,140 @@ pub fn modal(input: TokenStream) -> TokenStream {
...
+    let path = attr
+        .path()
+        .segments
+        .iter()
+        .map(|s| format!("{}", s.ident))
+        .collect::<Vec<String>>()
+        .join("::");
+
+    path == "poise::command" || path == "command"
+}
```
GnomedDev: Can't this be implemented without the formatting and the vector allocation and stuff? Just check the first and second segments?

GnomedDev (changes requested):
Looks okay, just a few nitpicks that would improve readability before I dive too deep into it.

macros/src/lib.rs
TitaniumBrain: Pretty sure, yeah. Editor added that for some reason.

macros/src/lib.rs
TitaniumBrain: This _is_ ugly.
What do you think of
```rust
if let Some(segment) = path.attr.path().segments.last() {
    segment.ident == "command"
} else {
    false
}
```
Edit: idk if Ident can be compared like that, I'm on mobile rn, but you get the point.

GnomedDev: That wouldn't handle the case of `bogus::command` correctly

TitaniumBrain: Neither would the previous version, tbf, the user could just
```rust
use poise::command;
// use bogus::command;

#[command(...)]
```

And, afaik, a macro can't know which is which.
I don't think it is s problem though, since unless `bogus::command` takes the same attributes as `poise::command`, it won't compile.

You could always only match `poise::command`, I guess.

macros/src/lib.rs
```diff
@@ -277,3 +279,117 @@ pub fn modal(input: TokenStream) -> TokenStream {
...
+
+fn group_impl(args: TokenStream, input_item: TokenStream) -> Result<TokenStream, Error> {
+    let args = darling::ast::NestedMeta::parse_meta_list(args.into())?;
+
+    let group_args = <command::GroupArgs as darling::FromMeta>::from_list(&args)?;
+
+    // let item_impl = syn::parse_macro_input!(input_item as syn::ItemImpl);
+    let item_impl = match syn::parse::<syn::ItemImpl>(input_item) {
+        Ok(syntax_tree) => syntax_tree,
+        Err(err) => return Err(err.into()),
```
GnomedDev, line 331: I have a feeling this is exactly what `?` does, it already calls `into`, but if this stops an error sure.

macros/src/lib.rs
```diff
@@ -277,3 +279,117 @@ pub fn modal(input: TokenStream) -> TokenStream {
...
+        if let syn::ImplItem::Fn(f) = item {
+            // ... and it's a command
+            if let Some(attr) = f.attrs.iter().find(|attr| is_command_attr(attr)) {
+                // add to command list
+                command_idents.push(f.sig.ident.clone());
+
+                // Turn a syn::Attribute into command::CommandArgs
+                let attr_args = &attr.meta.require_list()?.tokens;
+
+                let command_args = darling::ast::NestedMeta::parse_meta_list(quote! {#attr_args})?;
```
GnomedDev: Nitpick, feel free to ignore: `quote!(#whatever)` is just `whatever.to_token_stream()`

macros/src/lib.rs
```diff
@@ -277,3 +279,117 @@ pub fn modal(input: TokenStream) -> TokenStream {
...
+        }
+        impl_body = quote!(
+            #impl_body
+            #item_stream
+        );
+    }
+
+    Ok(quote! {
+        impl #name {
+            fn commands() -> Vec<Command<Data, Error>> {
```
GnomedDev: Where are these `Data` and `Error` idents coming from? Is this just assuming something is in scope with `Data` and `Error`? There is most likely a better way to handle this.

macros/src/lib.rs
```diff
@@ -277,3 +279,117 @@ pub fn modal(input: TokenStream) -> TokenStream {
...
+                vec![#(#name::#command_idents()),*]
```
GnomedDev: Nitpick, feel free to ignore: This could be a fixed length array, avoiding the heap alloc for minimal performance on startup which really doesn't matter.

macros/src/lib.rs
```diff
@@ -277,3 +279,117 @@ pub fn modal(input: TokenStream) -> TokenStream {
...
+
+            #impl_body
+        }
+    }
+    .into())
+}
+
+/**
+Returns true if an `Attribute` has `path` equal to "poise::command" or "command"
+*/
```
GnomedDev: Nitpick: For small comments like this, it's better to use `///` instead of `/* */`, I'm somewhat surprised that poise uses the second style at all.

macros/src/lib.rs
```diff
@@ -7,7 +7,9 @@ mod command;
 mod modal;
 mod util;
 
+use darling::Error;
```
GnomedDev: Probably better to refer to this using the path when used instead of importing like this, as there are a couple error types in play in this file.

macros/src/command/mod.rs
```diff
@@ -57,6 +57,226 @@ pub struct CommandArgs {
...
+///
+/// Same as CommandArgs, but with some removed, because they wouldn't make sense
+#[derive(Default, Debug, darling::FromMeta)]
+#[darling(default)]
+pub struct GroupArgs {
+    prefix_command: bool,
+    slash_command: bool,
+    context_menu_command: Option<String>,
+
+    // When changing these, document it in parent file!
```
GnomedDev, line 264: Are these actually documented anywhere?

macros/src/lib.rs
```diff
@@ -277,3 +279,117 @@ pub fn modal(input: TokenStream) -> TokenStream {
         Err(e) => e.write_errors().into(),
     }
 }
+
+/**
+Use this macro on an impl block to implement the CommandGroup trait.
```
GnomedDev, line 286: What `CommandGroup` trait?

GnomedDev (changes requested):

macros/src/lib.rs
TitaniumBrain: Just noticing I left that comment by accident. 
What I have is almost what the macro would expand into, but iirc it wouldn't return the right error type (return type may need work, a Box dyn or whatever)

macros/src/lib.rs
TitaniumBrain: Personally, I don't mind the macro since I know what it does, but I guess o can reserve that for when I'm building a stream now complex than just `#foo`.

Speaking of this, where I use an empty `quote!()` I should probably do `TokenStream::new()`

macros/src/lib.rs
TitaniumBrain: Yeah, that's what I did to please the compiler, since iirc from the examples, the user should define their own Data and Error.

Maybe these can be trait bounds?

macros/src/lib.rs
TitaniumBrain: Oh yeah 😅 that doesn't exist. Should I add it?

macros/src/command/mod.rs
TitaniumBrain: I copied that from `CommandArgs`, I'm not sure what it really refers to.

Edit: nvm, it's probably the doc comment on `command`.

I think it makes sense to leave that comment there, but I want to know how you prefer to document `group`:
- list all allowed attributes, which is copy pasting from `command` with a few removals; 
or
- list the differences between this and `command`, then add a "Refer to poise::command for a full list of attributes";

macros/src/lib.rs
GnomedDev: Combining this with the CommandGroup suggestion, CommandGroup could be generic over Data and Error, then this can be 
```rs
pub trait CommandGroup {
    type Data;
    type Error;
    fn commands() -> Vec<Command<Self::Data, Self::Error>>;
}

impl<D, E> CommandGroup for #name {
    type Data = D;
    type Error = E;
    fn commands() -> Vec<Command<D, E>>;
}
```

macros/src/lib.rs
GnomedDev: Yes, more details in another review comment.

macros/src/command/mod.rs
GnomedDev: The second option would be preferred

macros/src/lib.rs
```diff
@@ -361,6 +365,26 @@ fn group_impl(args: TokenStream, input_item: TokenStream) -> Result<TokenStream,
                     sig: f.sig.clone(),
                     block: Box::new(f.block.clone()),
                 };
+
+                if let None = ctx_type_with_static {
```
GnomedDev: why isn't this `if ctx_type_with_static.is_none()`

macros/src/lib.rs
```diff
@@ -361,6 +365,26 @@ fn group_impl(args: TokenStream, input_item: TokenStream) -> Result<TokenStream,
...
+                                "expected a Context parameter",
+                            )
+                            .into())
+                        }
+                    };
+                    // Needed because we're not allowed to have lifetimes in the hacky use case below (in command::mod.rs)
+                    ctx_type_with_static = Some(syn::fold::fold_type(
+                        &mut crate::util::AllLifetimesToStatic,
+                        context_type
+                            .expect("context_type has already been set to Some")
```
GnomedDev: Nitpick: `context_type should have already been set`

GnomedDev (commented):
LGTM, just a couple minor fixes and I'll merge

macros/src/lib.rs
```diff
@@ -277,3 +279,146 @@ pub fn modal(input: TokenStream) -> TokenStream {
...
+        Err(err) => err.write_errors().into(),
+    }
+}
+
+fn group_impl(args: TokenStream, input_item: TokenStream) -> Result<TokenStream, darling::Error> {
+    let args = darling::ast::NestedMeta::parse_meta_list(args.into())?;
+
+    let group_args = <command::GroupArgs as darling::FromMeta>::from_list(&args)?;
+
+    // let item_impl = syn::parse_macro_input!(input_item as syn::ItemImpl);
```
GnomedDev:
```suggestion

```

macros/src/lib.rs
```diff
@@ -277,3 +279,146 @@ pub fn modal(input: TokenStream) -> TokenStream {
...
+                let function = syn::ItemFn {
+                    attrs: vec![],
+                    vis: f.vis.clone(),
+                    sig: f.sig.clone(),
+                    block: Box::new(f.block.clone()),
+                };
+
+                if ctx_type_with_static.is_none() {
+                    let context_type = match function.sig.inputs.first() {
+                        Some(syn::FnArg::Typed(syn::PatType { ty, .. })) => Some(&**ty),
```
GnomedDev: Why return `Some` here just to unwrap it, there isn't a branch that returns None?

macros/src/lib.rs
```diff
@@ -277,3 +279,146 @@ pub fn modal(input: TokenStream) -> TokenStream {
...
+                }
+                item_stream = command::command(new_args, function)?.into();
+            }
+        }
+        impl_body = quote!(
+            #impl_body
+            #item_stream
+        );
+    }
+
```
GnomedDev, line 395: Handle `ctx_type_with_static` being none here?

examples/group_testing/main.rs
```diff
@@ -0,0 +1,116 @@
...
+impl Test {
+    // Just a test
+    #[poise::command(slash_command, prefix_command, rename = "test")]
+    async fn test_command(ctx: Context<'_>) -> Result<(), Error> {
+        let name = ctx.author();
+        ctx.say(format!("Hello, {}", name)).await?;
+        Ok(())
+    }
+}
+
```
GnomedDev, line 23: Can your add a test for multiple commands, and a test for no commands in a group?

GnomedDev (changes requested):

examples/group_testing/main.rs
TitaniumBrain: Just a couple structs, one with no commands and one with 2 or 3 in the same file?
Sounds good enough?

PR #269, opened 2024-05-20
Add support for using manual cooldowns on single commands

jamesbt365:
Not sure if adding fields to `Command` is breaking, it can be constructed publicly so maybe? will rebase to `next` on request.

Currently it is treated like an override, but if allowing users to have manual cooldowns on for every command but a few select ones is desired, this could be an `Option<bool>`.

I have not adjusted the examples, but in the current state there isn't really anything to add. Maybe a comment somewhere saying that it can be done for specific commands?

### DIFF
--- a/macros/src/command/mod.rs
+++ b/macros/src/command/mod.rs
@@ -49,7 +49,7 @@ pub struct CommandArgs {
     category: Option<String>,
     custom_data: Option<syn::Expr>,
 
-    manual_cooldowns: bool,
+    manual_cooldowns: Option<bool>,
 
     // In seconds
     global_cooldown: Option<u64>,
@@ -282,7 +282,7 @@ fn generate_command(mut inv: Invocation) -> Result<proc_macro2::TokenStream, dar
     let description = wrap_option_to_string(inv.description.as_ref());
     let category = wrap_option_to_string(inv.args.category.as_ref());
 
-    let manual_cooldowns = inv.args.manual_cooldowns;
+    let manual_cooldowns = wrap_option(inv.args.manual_cooldowns);
     let cooldown_config = generate_cooldown_config(&inv.args);
 
     let default_member_permissions = &inv.default_member_permissions;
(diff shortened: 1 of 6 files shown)


macros/src/command/mod.rs
```diff
@@ -49,6 +49,8 @@ pub struct CommandArgs {
     category: Option<String>,
     custom_data: Option<syn::Expr>,
 
+    manual_cooldowns: bool,
```
kangalio: Yeah I'd vouch for `Option<bool>`. Negligible extra complexity imo (especially not if you just use the attribute macro API, since it makes no difference there), but keeps the door open for more funky use cases users will think of

macros/src/command/slash.rs
```diff
@@ -215,7 +215,7 @@ pub fn generate_context_menu_action(
     Ok(quote::quote! {
         <#param_type as ::poise::ContextMenuParameter<_, _>>::to_action(|ctx, value| {
             Box::pin(async move {
-                if !ctx.framework.options.manual_cooldowns {
+                if !ctx.framework.options.manual_cooldowns && !ctx.command.manual_cooldowns {
```
kangalio: (nitpick, you may ignore to your own discretion) There's three of these `!ctx.framework.options.manual_cooldowns && !ctx.command.manual_cooldowns` now, maybe they can be centralized somewhere. So that it's harder to forget one of the two clauses in future code, if another one of these if's will be added in the future

kangalio (dismissed):

macros/src/command/slash.rs
jamesbt365: Would involve making a function public somewhere that would then allow users to interact with, not sure if that is favorable, the logic itself isn't that complex so I don't believe its necessary at this time.

macros/src/command/mod.rs
jamesbt365: Resolved.

GnomedDev (approved):

PR #270, opened 2024-05-20
Add support for modifying the last invocation of a command

jamesbt365:


src/cooldown.rs
```diff
@@ -122,6 +139,24 @@ impl CooldownTracker {
...
+            CooldownType::Guild(guild_id) => {
+                self.guild_invocations.insert(guild_id, instant);
+            }
+            CooldownType::Channel(channel_id) => {
+                self.channel_invocations.insert(channel_id, instant);
+            }
+            CooldownType::Member(member) => {
+                self.member_invocations.insert(member, instant);
+            }
+        }
```
kangalio: Reminds me of the "no boolean parameters" clean code rule, but with 5 execution paths instead of 2... I'd personally have made 5 individual functions instead

But this is nitpicky. I'm fine with this version too

kangalio (approved):

src/cooldown.rs
```diff
@@ -122,6 +139,24 @@ impl CooldownTracker {
             self.member_invocations.insert((ctx.user_id, guild_id), now);
         }
     }
+    /// Sets the last invocation for the specified cooldown bucket.
```
kangalio, line 143: Maybe add a sentence that hints that you usually don't need this, that this is more of a special purpose function

`This function is usually not needed for regular usage. It was added for more flexibility in exceptional situations" or something like that. Since you made the PR, you probably even have some specific example use case :)

src/cooldown.rs
jamesbt365: This was decided against in the end: https://canary.discord.com/channels/381880193251409931/1246193397950058809/1246385743144681502

src/cooldown.rs
jamesbt365: Pushed a new commit that does this.

kangalio (approved):

PR #271, opened 2024-05-21
Fix serenity next breakage due to dm-permission being removed in serenity next

supercoolspy:
Title


GnomedDev:
This removes the `guild_only` parameter entirely, which is not the solution and will break bots at runtime. I will fix this when I have time and energy, but for now you can pin to the serenity commit before the breakage. A reminder that this is what you get when using unstable branches, they break and have no guarantees for even compiling.

PR #277, opened 2024-05-26
Disable allowed mentions on certain errors

### LINKED ISSUE #263: ArgumentParse errors from poise::command are not sanitized
https://github.com/serenity-rs/poise/blob/186897a2f4f18b59152c9e670db5cd328bb22ea7/src/builtins/mod.rs#L86-L90

this can easily lead to many side effects with not-very-specially formatted messages like
> hello\` <@746501080409702461> `world

jamesbt365:
Fixes #263.

Disables mentions so errors don't break out and ping people, at least in errors that i reasonably suspect this could happen accidentally.

Not going to disable it on specific errors like `SubcommandRequired` because if you name your subcommand something that escapes the codeblock and pings people, thats on you.

If there is anything I missed I'll add it in, just mention the name of the error.

### DIFF
--- a/src/builtins/mod.rs
+++ b/src/builtins/mod.rs
@@ -17,7 +17,7 @@ mod paginate;
 #[cfg(any(feature = "chrono", feature = "time"))]
 pub use paginate::*;
 
-use crate::{serenity_prelude as serenity, CreateReply};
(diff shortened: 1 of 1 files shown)


NotNorom (approved):
LGTM

PR #279, opened 2024-05-28
Set edition from 2018 to 2021

NotNorom:
I mean... why not :D


GnomedDev:
This should be targeted at current, as edition is a crate-internal change and the MSRV is met on current.

NotNorom:
sure!

PR #282, opened 2024-06-08
Fix compile error

skyfallwastaken:
Before this commit, Poise compilation would fail with this error:

```
error[E0432]: unresolved import `crate::serenity`
  --> src/builtins/mod.rs:20:13
   |
20 | use crate::{serenity::CreateAllowedMentions, serenity_prelude as serenity, CreateReply};
   |             ^^^^^^^^
   |             |
   |             unresolved import
   |             help: a similar path exists: `crate::prefix_argument::key_value_args::serenity`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `poise` (lib) due to 1 previous error
```

Now, compilation succeeds successfully.


skyfallwastaken:
looks like `serenity_context` also got renamed, which is weird because it didn't show up on my cargo clippy 😅

Anyway I'll probably try to fix most of those later

skyfallwastaken:
Okay, so I _think_ that's everything! The tests are working anyways

GnomedDev:
Thanks for the PR, however I've fixed this up via rebasing and editing the specific commits that brought in the issues.

PR #286, opened 2024-07-15
Rework permission checks

jamesbt365:
This shouldn't be breaking in the traditional sense, but does theoretically change behaviour in one case.

Previously, when the bots permissions could not be checked, it let the command run, now it fires a UserPermissionsError, there isn't really a nice way of checking where the error came from because the only place this can happen is in http requests.

The guild, channel or member itself could be the failure point, the only place we can now tell is directly on which member failed to fetch and return the "right" error from there.

I've placed a comment in the code about it, I just want to check if this is okay or if I should attempt to get the right error even though the only place it can come from is http.

I can pull the member from the cache too if that is desirable, but developers without `GUILD_MEMBERS` may suffer from this change. I believe checking for this is doable by directly accessing `Http`, but this is currently out of scope and can be handled later if that is desirable.

In `serenity-next` the channel http request has already been removed and I'd rather not mess around doing that bit manually and falling back to http if it isn't cached.

CURRENTLY UNTESTED.

### DIFF
--- a/src/dispatch/common.rs
+++ b/src/dispatch/common.rs
@@ -2,23 +2,45 @@
 
 use crate::serenity_prelude as serenity;
 
+/// Used in `user_permissions`, if a user is passed to that function, their permissions will be
+/// returned, in DMs the return value with be `Permissions::all()`.
+struct PermissionsInfo {
+    /// The Permissions of the user, if requested.
+    user_permissions: Option<serenity::Permissions>,
+    /// The Permissions of the bot, if requested.
+    bot_permissions: Option<serenity::Permissions>,
+}
+
 /// Retrieves user permissions in the given channel. If unknown, returns None. If in DMs, returns
 /// `Permissions::all()`.
-async fn user_permissions(
-    ctx: &serenity::Context,
-    guild_id: Option<serenity::GuildId>,
-    channel_id: serenity::ChannelId,
-    user_id: serenity::UserId,
-) -> Option<serenity::Permissions> {
-    let guild_id = match guild_id {
-        Some(x) => x,
-        None => return Some(serenity::Permissions::all()), // no permission checks in DMs
+async fn users_permissions<U, E>(
+    ctx: crate::Context<'_, U, E>,
+    user_id: Option<serenity::UserId>,
+    bot_id: Option<serenity::UserId>,
+) -> Option<PermissionsInfo> {
+    let guild_id = ctx.guild_id();
+    let channel_id = ctx.channel_id();
+    // No permission checks in DMs.
+    let Some(guild_id) = guild_id else {
+        return Some(PermissionsInfo {
+            user_permissions: Some(serenity::Permissions::all()),
+            bot_permissions: Some(serenity::Permissions::all()),
+        });
     };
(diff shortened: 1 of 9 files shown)


GnomedDev:
Okay, currently this code is a mess because of cache and no cache support, and I see you adding the signature `Option<(Option<_>, Option<_>)>` which is sus to start with.

Can you abstract the distinct operations that need difference between cache and http into functions in a different file that we can switch between?

jamesbt365:
Lets go from where the most recent commit is, what else ideally should be done to make this more presentable and maintainable?

src/dispatch/common.rs
```diff
@@ -2,23 +2,48 @@
...
+            // funny sometimes, so lets be safe.
+            if let Some(member) = &ctx.interaction.member {
+                return Some(PermissionsInfo {
+                    user_permissions: member.permissions,
+                    bot_permissions: ctx.interaction.app_permissions,
+                });
+            }
+        }
+        crate::Context::Prefix(_) => {}
+    }
```
GnomedDev: `if let`?

src/dispatch/common.rs
```diff
@@ -29,31 +54,124 @@ async fn user_permissions(
...
+        return fetch_guild(ctx, channel, user, bot).await;
+    };
+
+    #[cfg(feature = "cache")]
+    return Some(permissions);
 
-    Some(guild.user_permissions_in(&channel, &member))
+    #[cfg(not(feature = "cache"))]
+    fetch_guild(ctx, channel, user, bot).await
+}
```
GnomedDev: `if let`, inside the body return, then fallback to `fetch_guild`. That will remove some of the cfg attributes.

src/dispatch/common.rs
```diff
@@ -111,35 +229,38 @@ async fn check_permissions_and_cooldown_single<'a, U, E>(
...
+
+        // missing premission checks here.
+    } else {
+        // TODO: ask what I should do here because combining the checks loses the verbosity.
+        // the only previous failure point was it failing to get the guild, channel or members.
+        // Previously when a bots permissions could not be fetched it would just allow execution.
+        return Err(crate::FrameworkError::MissingUserPermissions {
+            missing_permissions: None,
+            ctx,
+        });
```
GnomedDev: My only thought is "swap to next, add new variant".

GnomedDev (commented):
Okay, this looks a bit better (may just be me getting some sleep). Just a few improvements that will make this even better, and some testing needs to be done if not already.

src/dispatch/common.rs
jamesbt365: Swap to a generic MissingPermissions?

GnomedDev: CannotFetchPermissions but with a better name.

GnomedDev: This still isn't done.

jamesbt365: Oops, i added the variant but didn't actually use it, should be fixed in the latest commit.

GnomedDev:
Closing in favor of #325.

PR #297, opened 2024-08-11
Fix default_member_permissions not being checked for context_menu_command

zkxs:
This PR fixes default_member_permissions not being checked for context menu commands.

Commands with `context_menu_command` currently do not use the command macro's configured default_member_permissions. If default_member_permissions is not set in the command's builder, the context menu command will default to allowing all users.

This is a security footgun, as it's very easy to make a command intended for privileged users and overlook the fact that the permissions are only enforced on the slash command variant and not the context menu command. My bot exclusively uses default_member_permissions for command permission checks and I almost missed this bug, which would have given all users access to commands intended only for server moderators.

### DIFF
--- a/src/structs/command.rs
+++ b/src/structs/command.rs
@@ -249,8 +249,7 @@ impl<U, E> Command<U, E> {
         });
 
         // This is_empty check is needed because Discord special cases empty
-        // default_member_permissions to mean "everyone". Without this, all Poise context menu commands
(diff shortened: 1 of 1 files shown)


zkxs:
rebased onto latest `current` branch (this PR has been idle for 5 months so it probably needed it).

meditationmind:
Just tested this and everything works as expected!

arqunis (approved):

PR #298, opened 2024-08-12
create_application_commands can now take any IntoIterator: not just &[]

zkxs:
This is a small change: currently `poise::builtins::create_application_commands` takes a `&[crate::Command<U, E>]`. This PR changes it to take an `IntoIterator<Item=&crate::Command<U, E>>`. This doesn't really change much, except it makes the API slightly easier to work with if API users don't simply have a `Vec<Command<U, E>` lying around: for example if you're dynamically computing which commands should be set on a guild-by-guild basis.

### DIFF
--- a/src/builtins/register.rs
+++ b/src/builtins/register.rs
@@ -16,8 +16,8 @@ use crate::serenity_prelude as serenity;
 /// serenity::Command::set_global_commands(ctx, create_commands).await?;
 /// # Ok(()) }
 /// ```
-pub fn create_application_commands<U, E>(
-    commands: &[crate::Command<U, E>],
+pub fn create_application_commands<'a, U:'a , E: 'a>(
+    commands: impl IntoIterator<Item=&'a crate::Command<U, E>>,
 ) -> Vec<serenity::CreateCommand> {
     /// We decided to extract context menu commands recursively, despite the subcommand hierarchy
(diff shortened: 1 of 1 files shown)


GnomedDev:
This is a breaking change, therefore needs to be targeted at the `next` branch.

zkxs:
@GnomedDev after some confusion between the `serenity-next` and `next` branches I've changed the base.

GnomedDev:
Do you want to also update the functions that call create_application_commands to take advantage of the more permissive function signature?

zkxs:
Sure; it should be done now unless I missed any: it looks like it's just `register_in_guild` and `register_globally`.

GnomedDev:
Ignore the clippy issues, they are failing for unrelated issues, while you fix the `rustfmt` issue could you use `Iterator::size_hint` to restore the `Vec::with_capacity` optimization?

zkxs:
@GnomedDev Should be done. I'm not checking the upper bound on size_hint, but that seems correct to me because in my experience overestimating the allocation tends to hurt performance more than underestimating.

zkxs:
rebased onto latest `next` branch (this PR has been idle for 5 months so it probably needed it).

NotNorom:
LGTM

arqunis (approved):

PR #301, opened 2024-09-11
Allow edited messages and interactions to be sent without removing existing embeds

### LINKED ISSUE #300: Reply edit without reset embed
Why embeds on reply edit always replaced and not made as `allowed_mentions` for example?
Because of that I can't just edit components without not cloning previous embeds.

https://github.com/serenity-rs/poise/blob/575025909b063c3b998659abf9d241c8790404ee/src/reply/builder.rs#L174-L178

UPD.
Okay, I found that for `CreateReply` field `embeds` is not `Option`. Why?

TapGhoul:
Allows replies to be sent with embeds nulled out in the API, which prevents removing existing embeds under the context of editing an existing message response, both for interactions and standard messages.

This changes existing behavior - now, when not setting any embeds, existing embeds are kept. To clear them, you would call `reply.embeds(vec![])`

Pairs with https://github.com/serenity-rs/serenity/pull/2968 - this change can be made without it, but together they fix followup messages too. They can be merged separately, though the documentation on the `embed()` method is considered incorrect until this is merged on Serenity's side.

Fixes #300

### DIFF
--- a/src/builtins/pretty_help.rs
+++ b/src/builtins/pretty_help.rs
@@ -36,7 +36,7 @@ impl Default for PrettyHelpConfiguration<'_> {
     }
 }
 
-/// A help command that works similarly to `builtin::help` butt outputs text in an embed.
+/// A help command that works similarly to `builtin::help` but outputs text in an embed.
 ///
 pub async fn pretty_help<U, E>(
     ctx: crate::Context<'_, U, E>,
--- a/src/reply/builder.rs
+++ b/src/reply/builder.rs
@@ -32,7 +32,7 @@ impl CreateReply {
     /// Existing embeds on this are kept.
     /// When editing a message, this will overwrite previously sent embeds.
     pub fn embed(mut self, embed: serenity::CreateEmbed) -> Self {
-        self.embeds.get_or_insert_with(|| Default::default()).push(embed);
+        self.embeds.get_or_insert_with(Vec::new).push(embed);
         self
     }
 
--- a/src/reply/builder.rs
+++ b/src/reply/builder.rs
@@ -114,11 +114,11 @@ impl CreateReply {
         if let Some(components) = components {
             builder = builder.components(components);
         }
-        if let Some(ephemeral) = ephemeral {
-            builder = builder.ephemeral(ephemeral);
-        }
         if let Some(embeds) = embeds {
             builder = builder.embeds(embeds);
+        }
+        if let Some(ephemeral) = ephemeral {
+            builder = builder.ephemeral(ephemeral);
         }
 
         builder.add_files(attachments)
@@ -142,11 +142,11 @@ impl CreateReply {
         if let Some(content) = content {
             builder = builder.content(content);
         }
-        if let Some(embeds) = embeds {
-            builder = builder.embeds(embeds);
-        }
         if let Some(components) = components {
             builder = builder.components(components)
+        }
+        if let Some(embeds) = embeds {
+            builder = builder.embeds(embeds);
         }
         if let Some(allowed_mentions) = allowed_mentions {
             builder = builder.allowed_mentions(allowed_mentions);
@@ -179,11 +179,11 @@ impl CreateReply {
         if let Some(components) = components {
             builder = builder.components(components);
         }
-        if let Some(allowed_mentions) = allowed_mentions {
-            builder = builder.allowed_mentions(allowed_mentions);
-        }
         if let Some(embeds) = embeds {
             builder = builder.embeds(embeds);
+        }
+        if let Some(allowed_mentions) = allowed_mentions {
+            builder = builder.allowed_mentions(allowed_mentions);
         }
 
         builder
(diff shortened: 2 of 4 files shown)


jamesbt365:
This is a breaking change because:
- `CreateReply`'s fields are public and you changed them
- It changes the expected behaviour for users requiring them to make code changes on minor version bumps.

Target `next`.

Additionally, the current behaviour is apparently more intuitive, and you still have access to the old embeds when handling a ReplyHandle, so its not like the current way is impossible to work with.
https://github.com/serenity-rs/poise/blob/575025909b063c3b998659abf9d241c8790404ee/src/reply/mod.rs#L107-L111

The only thing with any mild annoyance is that there isn't a nice way to keep attachments without reuploading them like the serenity builders?

TapGhoul:
Oop! My bad on the public field breaking change, did this very early in the morning. All good!

The code you highlighted applies to replies, not interactions. Interactions force clear embeds every time, hence the referenced PR in Serenity core too, as this is incorrect on their side as well, which was my focus. However, by force-setting embeds every time, you actually explicitly remove all embeds, unless you supply the full embeds list again - and this is not supplied in the edit/reply builder.

-----

As for the embeds, feel free to correct me if I'm wrong as I'm just doing a code walk, not running tests against Poise itself (though I created the pair of PRs out of observed misbehavior, and tested against the API directly) - but from what I understand that's not how that code works, and the comment is outdated.

- To start off with, you construct CreateReply with `CreateReply::default()` - there is no way to go from a message to a builder that I can find. So no prefill here.
- For all branches (for reply, edit interaction initial and edit interaction followup) Poise first takes in the `CreateReply` struct, and modify it by applying `ephimeral` and `allow_mentions` set as per the command + framework options if not already set. All good thus far.
- Pose then runs the conversion methods (`to_prefix_edit()`, `to_slash_initial_response_edit()` and `to_slash_followup_response()`) with an empty response and tack on what's needed. Fine so far.

Now here's where things go wrong: When generating these, you then always explicitly set `.embeds()` - on `EditMessage`, this is where embeds get forcefully cleared. *There is no way for the discord API to append new embeds to an existing list of embeds on a message,* only set or leave them - I believe there was once some old behavior that did this, but this hasn't been the case in a long time. But your code explicitly removes embeds each time, rather than allowing them to remain.

In the case of interactions, it gets worse - while the same behavior as above applies, serenity itself steps on your toes itself in the case of editing the followup interaction responses (though not the initial one). Even if you prevent this, serenity made the embeds field mandatory by mistake. The serenity-side PR mentioned in my initial comment fixes that, and a similar fix was merged previously for the `CreateInteractionResponseData` call (now called `CreateInteractionResponseMessage`) - https://github.com/serenity-rs/serenity/pull/2216.

Poise use this in your own pagination builtin, when pagination occurs on the initial message, if triggered by an interaction such as a slash command. You don't see any issues with yours because you rebuild the embed for every page, but in situations where this is not the case, this breaks.


The test code used to perform this in verifying my claims in this comment, using xh (replace with curl or your favourite HTTP API testing software):
```sh
send() {
	xh $1 "https://discord.com/api/v10/$2" Authorization:'Bot <token>' User-Agent:'DiscordBot (<url>, 0.0.1)' ${@:3}
}

# Create the message, take note of the ID
send POST channels/<channelid>/messages content=hi embeds:='[{"title":"hi"}]'
# This will edit the message, but not remove embeds
send PATCH channels/<channelid>/messages/<msgid> content=hello
# This will remove the existing embed, and replace it with the one given here.
send PATCH channels/<channelid>/messages/<msgid> content=hello embeds:='[{"title":"hello"}]'
# This will remove all embeds, and is what Poise does by default
send PATCH channels/<channelid>/messages/<msgid> content=hello embeds:='[]'
```

jamesbt365:
> The code you highlighted applies to replies, not interactions.

`ReplyHandle` is a wrapper around the bot's interaction responses and the bot's prefix responses. It is returned from poise's methods on `Context` and is primarily used to *edit* or delete responses.

Am I missing anything here? You mention editing the response which `ReplyHandle` is used for.

TapGhoul:
When calling `handle.edit()`, you feed in a builder you construct yourself. It does not give you a pre-configured builder. On build inside edit, it converts it into the underlying Serenity type, and uses the serenity message/interaction methods to fill in the message ID and other context held within the handle's inner (as opposed to Poise managing that itself).

`ReplyHandle` is fine (though has an outdated comment), `ReplyBuilder` is the one with issues, and does not get constructed from `ReplyHandle` directly. It is instead fed into the edit method, already hand-assembled by the user.

TapGhoul:
Trying to get it properly on poise next. Will reopen once I get it sorted.

jamesbt365:
Take this as a lesson to stop opening PRs from the main branch of your fork.

TapGhoul:
I mean, that's the flow github offers up now. Used to not, used to autogenerate a branch called patch-1. I guess something changed. Either way, opened #302.

PR #302, opened 2024-09-11
(Cont #301) Allow edited messages and interactions to be sent without removing existing embeds

TapGhoul:
Continuing from #301


TapGhoul:
CC: @jamesbt365

PR #318, opened 2024-10-31
Swap Command and related structs to avoid String allocs

GnomedDev:
This avoids allocating a bunch of `String`s and `Vec`s when they can just be literals. This significantly simplifies command construction functions towards the goal of making them `static`, or close to at least.

This does not impact modifying these values at runtime as `Cow<'static, str>` can store `String` as well as `&'static str`.


jamesbt365 (approved):
If it works at runtime it LGTM

PR #320, opened 2024-11-01
Get rid of autoref-specialisation

GnomedDev:
This significantly simplifies poise internals, by removing the fancy autoderef specialization used to fallback from the poise traits, to the serenity traits, and finally to FromStr.

It is replaced by normal implementations, which does not significantly impact usability as users may define newtypes to implement the traits themselves if needed.

This PR is currently missing:
- testing to make sure I didn't break anything


mkrasnitski:
What sort of runtime testing is required for this?

jamesbt365:
> What sort of runtime testing is required for this?

I'm guessing just testing if all the current things that parse still parse the same way, check the existing impls and cross reference i guess?

If that all works it should be fine

mkrasnitski:
I'm not even sure what the specialization was trying to solve. Was it a workaround for trying to write a blanket implementation of `impl<T: FromStr> SlashArgument for T` in order to avoid overlapping impls (and likewise for `PopArgument`)?

GnomedDev:
Yes.

macros/src/command/slash.rs
```diff
@@ -80,7 +80,7 @@ pub fn generate_parameters(inv: &Invocation) -> Result<Vec<proc_macro2::TokenStr
                     quote::quote! { Some(|o| o.kind(::poise::serenity_prelude::CommandOptionType::Integer)) }
                 } else {
                     quote::quote! { Some(|o| {
-                        poise::create_slash_argument!(#type_, o)
+                        <#type_ as poise::SlashArgument>::create(o)
```
mkrasnitski, line 83:
```suggestion
                        <#type_ as ::poise::SlashArgument>::create(o)
```
Minor nit

src/prefix_argument/argument_trait.rs
```diff
@@ -1,106 +1,60 @@
...
+//! Trait implemented for all types usable as prefix command parameters.
+//!
+//! Many of these implementations defer to [`serenity::ArgumentConvert`].
 
 use super::{pop_string, InvalidBool, MissingAttachment, TooFewArguments};
 use crate::serenity_prelude as serenity;
-use std::marker::PhantomData;
 
-/// Full version of [`crate::PopArgument::pop_from`].
+/// The result of `<T as PopArgument>::pop_from`.
```
mkrasnitski, line 8:
```suggestion
/// The result of [`PopArgument::pop_from`].
```
Should keep the doc link.

src/prefix_argument/argument_trait.rs
```diff
@@ -1,106 +1,60 @@
...
-            .map_err(|e| (e.into(), Some(string)))?;
-
-        Ok((args.trim_start(), attachment_index, object))
-    }
+    ) -> PopArgumentResult<'a, Self>;
 }
 
 #[async_trait::async_trait]
-impl<'a, T: PopArgument<'a> + Send + Sync> PopArgumentHack<'a, T> for &PhantomData<T> {
+impl<'a> PopArgument<'a> for String {
```
mkrasnitski, line 36: Just noting the behavioral change here: string arguments now have leading whitespace trimmed, whereas it seems like that wasn't the case before. This is because `String` used the blanket `PopArgument` implementation for any `T: ArgumentConvert`, which in turn used the `T: FromStr` blanket impl for `ArgumentConvert`.

src/prefix_argument/argument_trait.rs
```diff
@@ -136,56 +86,52 @@ impl<'a> PopArgumentHack<'a, serenity::Attachment> for &PhantomData<serenity::At
...
+    u8, u16, u32, u64,
+    i8, i16, i32, i64,
+    serenity::UserId, serenity::User, serenity::Member,
+    serenity::MessageId, serenity::Message,
+    serenity::ChannelId, serenity::Channel, serenity::GuildChannel,
+    serenity::EmojiId, serenity::Emoji,
+    serenity::RoleId, serenity::Role
+);
+#[cfg(feature = "cache")]
+impl_popargument_via_argumentconvert!(serenity::GuildId, serenity::Guild);
```
mkrasnitski, line 137: I'm guessing these extra implementing types just come out of the aforementioned blanket impls of `T: ArgumentConvert` and subsequently `T: FromStr`.

There's a behavioral change here in that we're no longer parsing mentions of users, channels, or roles. Rather we're just parsing the raw snowflakes. Since `serenity::Mention` has a `FromStr` implementation that works for all three types, do you think we should add that to the list?

mkrasnitski (commented):
Overall this looks pretty good. Having to wrap an argument in `StrArg` to destructure it isn't the best and it would be nice to introduce an attribute that could do the specialization automatically; that can be done later as a followup. I noticed a couple behavioral changes, so if they're intentional then I'll approve.

src/prefix_argument/argument_trait.rs
jamesbt365: Doesn't this implementation remove the ping support for the Id types which was done via snowflake_pop_argument?

mkrasnitski: Which implementation are you referring to, do you mean `impl FromStr for Mention`? That one uses the same functions that `snowflake_pop_argument` used, as seen here: https://github.com/serenity-rs/serenity/blob/current/src/model/mention.rs#L135

jamesbt365: This, which allows use of the id types but users can specify mentions, isn't it just id parsing again now, not allowed to use mentions?

https://github.com/serenity-rs/poise/blob/a1edcc09085be800a8593602910d2a09eaf986dc/src/prefix_argument/argument_trait.rs#L139-L195

jamesbt365:
Tested this out a bit and it looks great other than the review comment, haven't had any other issues.

src/prefix_argument/argument_trait.rs
mkrasnitski: Yes, that's what I was referring to. Since we no longer pass `parse_{user,channel,role}_mention` to the macro, then IDs are just parsed as snowflakes, not as mentions.

mkrasnitski:
> Tested this out a bit and it looks great other than the review comment, haven't had any other issues.

Can you test to verify the behavior change around string arguments, where whitespace is now trimmed?

src/prefix_argument/argument_trait.rs
mkrasnitski: After re-examining this, the blanket `PopArgument` impl also used `pop_string` which is responsible for trimming leading whitespace. Therefore, I don't think there's a behavioral change here. It's worth testing though.

arqunis:
Superseded by https://github.com/serenity-rs/poise/pull/357

PR #324, opened 2024-11-03
Add poll support

jamesbt365:
This adds support for polls into poise. Due to these branches relying on crates.io releases and not git builds, this will not compile until the next release of serenity. Because of that this is marked as draft.

(provided I got my patches right) you should be able to use this *now* by doing this once serenity-rs/serenity#3013 is merged (or you can just point at that branch on my fork instead)

```toml
# you can change these to my forks if the PRs aren't merged.
poise = { git = "https://github.com/serenity-rs/poise", branch = "current"}
```

### DIFF
--- a/src/reply/builder.rs
+++ b/src/reply/builder.rs
@@ -180,7 +180,8 @@ impl CreateReply {
             components,
             ephemeral: _, // can't edit ephemerality in retrospect
             allowed_mentions,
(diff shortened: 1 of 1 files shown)


sawa-ko:
The pr mentioned is already merged.

GnomedDev (approved):

PR #336, opened 2025-01-27
fix MessageUpdateEvent breakage

jamesbt365:


GnomedDev:
Is this not just `Message::clone_from(update)`?

jamesbt365:
> Is this not just `Message::clone_from(update)`?

I guess but if poise was only using those fields this is cheaper? (though it was cloning the entire message lol)

I can use that instead if thats preferable.

GnomedDev:
Thanks for the PR, but I'm going to solve this slightly differently.

PR #340, opened 2025-03-02
WIP fixing breaking change on poise (serenity-next branch)

FabseGP:
Fixed accessing the ShardManager runners, but... don't know how to tackle 

```Rust
error[E0507]: cannot move out of `client.shard_manager` which is behind a shared reference
  --> src/framework/mod.rs:97:44
   |
97 |         self.shard_manager = Some(Arc::new(client.shard_manager));
   |                                            ^^^^^^^^^^^^^^^^^^^^ move occurs because `client.shard_manager` has type `serenity::gateway::ShardManager`, which does not implement the `Copy` trait
```
   
Serenity doesn't store shard_manager as an Arc, nor do shard_manager implement Copy or Clone (got changed in Serenity next-branch)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -114,9 +114,9 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bitflags"
-version = "2.8.0"
+version = "2.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8f68f53c83ab957f72c32642f3868eec03eb974d1fb82e453128456482613d36"
+checksum = "5c8214115b7bf84099f1309324e63141d4c5d7cc26862f97a0a857dbefe165bd"
 
 [[package]]
 name = "block-buffer"
@@ -169,9 +169,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.14"
+version = "1.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c3d1b2e905a3a7b00a6141adb0e4c0bb941d11caf55349d863942a1cc44e3c9"
+checksum = "be714c154be609ec7f5dad223a33bf1482fff90472de28f7362806e6d4832b8c"
 dependencies = [
  "shlex",
 ]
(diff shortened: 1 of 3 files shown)


yehorovye:
You cannot Arc shard manager as it doesn't implement Copy

FabseGP:
> You cannot Arc shard manager as it doesn't implement Copy

That makes sense, but... 

```Rust
    /// Initialized to Some during construction; so shouldn't be None at any observable point
    shard_manager: Option<Arc<serenity::ShardManager>>,
```

the framework expects an Arc for shard_manager. I did try and remove the Arc line by line, but I ended with a breakage in src/dispatch/mod.rs:

```Rust
error[E0308]: mismatched types
  --> src/dispatch/mod.rs:58:9
   |
57 |     pub fn shard_manager(&self) -> serenity::ShardManager {
   |                                    ---------------------- expected `serenity::gateway::ShardManager` because of return type
58 |         self.shard_manager.clone()
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `ShardManager`, found `&ShardManager`
   |
note: `serenity::gateway::ShardManager` does not implement `Clone`, so `&serenity::gateway::ShardManager` was cloned instead
  --> src/dispatch/mod.rs:58:9
   |
58 |         self.shard_manager.clone()
   |         ^^^^^^^^^^^^^^^^^^
``` 

(used to return Arc\<serenity::ShardManager\>)

More or less I'm stuck xD

FabseGP:
currently I'm at adding a Mutex to shard_manager in Serenity [(fork)](https://github.com/FabseGP/serenity-fork/commit/8917c0c3e3c0354ba533c5d6fb691925374b89d0) and then passing Arc<Mutex<ShardManager>> around, but that requires changing more than I like in Poise's framework

(as to why not just change it back to Arc<ShardManager>, shard_manager now have methods that require interior mutability)

cheesycod:
> currently I'm at adding a Mutex to shard_manager in Serenity [(fork)](https://github.com/FabseGP/serenity-fork/commit/8917c0c3e3c0354ba533c5d6fb691925374b89d0) and then passing Arc<Mutex> around, but that requires changing more than I like in Poise's framework
> 
> (as to why not just change it back to Arc, shard_manager now have methods that require interior mutability)

Checked your fork. The mutex looks highly suspicious though and might cause deadlocks if code ever tries to lock the shard manager after starting the bot?

jamesbt365:
Superseded by #341.

PR #345, opened 2025-04-14
Fix command macro not working under #![no_implicit_prelude]

abdullahqusim:
Simply modifies macro output to generate fully qualified paths such as `::core::option::Option::Some` instead of `Some` as recommended in macros, allowing the use of `#[command]` in crates with the attribute #![no_implicit_prelude] without missing import errors.


GnomedDev:
Can you add a test that this doesn't get broken over time?

PR #355, opened 2025-07-04
Improve edit tracking by adding options for old, uncached or suppression toggled messages

### LINKED ISSUE #309: Add more granual control to edit tracking
Currently, any message that is edited (regardless of how old) is tracked if edit tracking is enabled. This can lead to users editing very old messages and it invoking a bot command.

jamesbt365:
Fixes #309.

Currently untested at time of posting, but I want this looked at ASAP as currently untracked edits that receive message edits from discord (old message is reloaded with embeds) can cause commands to be executed again, these additional controls added by this PR should prevent this by using sane defaults.

But this is a big problem depending on the bot as this could cause roles to be added/removed from users, bans from being done/undone and other dangerous actions, this has been a problem since 2 criteria have been met:

1. Discord now sends the full message when a message is edited
2. Old messages now get message edit events with a client fetches the messages.

Added an additional control unrelated to this PR about message embed suppression though, as i thought about this when fixing this and i can see how this could be annoying to some.


arqunis:
Superseded by #402

PR #357, opened 2025-07-09
Remove specialization hacks for `PopArgument` and `SlashArgument`

mkrasnitski:
Supersedes #320. Replaces hacky blanket `impl {Pop,Slash}Argument for T` whenever `T: ArgumentConvert` and `T: FromStr` with explicit implementations generated via `macro_rules!`. One breaking change for `PopArgument` is that types which do not implement `ArgumentConvert` but implement `FromStr` must be decorated with a new `#[string]` attribute in order to generate the required parsing logic. Either that, or the user must implement the trait manually.

We also keep the current logic for parsing IDs out of @-mentions.

### DIFF
--- a/macros/src/command/mod.rs
+++ b/macros/src/command/mod.rs
@@ -79,6 +79,7 @@ struct ParamArgs {
     max: Option<syn::Lit>,
     min_length: Option<syn::Lit>,
     max_length: Option<syn::Lit>,
+    string: bool,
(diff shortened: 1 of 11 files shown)


arqunis (approved):

mkrasnitski:
Additional clippy failures are unrelated, so this PR is ready for merge.

Issue #313, opened 2024-10-22
Tracking Issue for poise-in-serenity


GnomedDev:
We are currently trying to merge poise into serenity, to replace the standard framework for 0.13+. This entails quite a bit of rework to make sure that the state of poise matches serenity. This is the issue to track that effort.

# Goals
- Remove as much of poise that is not a command framework as possible
- Remove unmaintained code, or split it off
- Remove as much magic as possible

# Progress
- [x] Remove user data setup (done, serenity-next)
- [x] Remove autocomplete magic (done, next, #312)
- [ ] Audit and cleanup `Command` (allocations, etc)
- [ ] Audit and cleanup `FrameworkOptions`
- [ ] Investigate the Modal derive, and if that should be split off
- [ ] Remove autoref specialisation for command handling, just use traits like everyone else (#320)
- [x] Remove help command infrastructure, should be split out into a separate crate (#323)
- - [ ] Investigate the rest of the builtins

PR #364, opened 2025-09-10
Make error replies ephemeral

HactarCE:
This might remove the need for #277.

Feel free to reject if there's a particular reason for these to be non-ephemeral.


SabrinaJewson:
Any update on whether this could be merged? This would be very useful.

meditationmind:
Considering that builtins are supposed to be a starting point, I could understand making an argument for or against ephemeral throughout, but I don't see the logic of mixing the two, or at least the current mix does not seem to be logical. Probably just a result of incremental changes over the years.

There are other inconsistencies and this function could probably use some overall cleanup/investigation, which is part of #313, but this PR seems fine as part of that process.

PR #377, opened 2026-01-13
Add new components to modal derive

meditationmind:
Derived modals currently only support text input components. This PR adds support for all components currently available to modals, with most of their functionality intact.

While I have tested the PR and everything works fine, I do have some questions as a first-time contributor:

1. I added some simple form validation checks, e.g., min/max value validity, making sure there aren't more than five components used in a single modal, etc. This is convenient because the errors come from the IDE while coding instead of from the API during runtime. Is this OK? Or do we prefer to leave form validation to the API so that we don't have to keep track of requirement changes?
2. I tried to use as much resolved data as possible, e.g., instead of returning `UserId`, I've returned `User`. Is this sensible? Also, as far as I can tell, resolved data for user select and mentionable select components always includes `PartialMember` along with `User`, so I've returned that in the `member` field. Is this also sensible?
3. Is `ModalDataResolved` an acceptable name for the struct that returns the resolved data for selected options? I based it on the existing `CommandDataResolved`. However, since this is only used in the derive, would `DerivedModalDataResolved` be better? More accurate, but a bit long.
4. ~~Would returning `FixedArray`s and `FixedString`s instead of `Vec`s and `String`s in `ModalDataResolved` be preferred?~~ Types returned via `ModalDataResolved` now align with Serenity types (`FixedArray`/`FixedString`).

If everything seems fine as is, then this can be merged without addressing the above questions. Thanks!


meditationmind:
Cleaned some things up and added the three newest components (radio group, checkbox group, and checkbox).

Still curious about the initial questions, but `#4`, in particular. Happy to switch to `FixedArray`/`FixedString` if that is preferred.

nwerosama:
I think `#4` is possible, I would've used `FixedString`/`FixedArray` if it was preferred and commonly used in other places but would probably confirm with another maintainer on this part.

I think the validation happens on API/Discord client side when it comes to min/max values and other things, so I suppose we can leave it up to Discord.

meditationmind:
> I think `#4` is possible

Certainly possible; I had already implemented it in stashed changes. I hesitated to deploy because these types seem less internal, since it's how the data is passed back to the user/dev, but assuming Poise will still eventually be integrated into Serenity, it probably makes the most sense. Lost the stashed changes, but will reimplement.

> I think the validation happens on API/Discord client side when it comes to min/max values and other things, so I suppose we can leave it up to Discord.

Dropping the validation makes the derive much less user-friendly, which is a big part of what Poise offers, IMO. With the validation, the dev gets immediate feedback while creating the struct. Without, you don't know anything is wrong until much later, while testing the bot in a live environment. Since the restrictions are documented, including them seems reasonable... was my line of thinking. Will reconsider, though, and hopefully get some more opinions. Appreciate the feedback.

meditationmind:
While reimplementing `FixedArray`/`FixedString`, I realized it makes the most sense to just return the same types as Serenity, since the derive is just an alternative method of using components from Serenity.

Some of the types could be further optimized, but that should be done in Serenity first, if at all. I have left non-doc comments specifying said optimizations on the `ModalDataResolved` struct.

meditationmind:
Had been a while so I did a self-review, cleaned up some code and made some other small improvements. Also added the new file types filter to the derive.

arqunis (approved):
Generally, we'd leave validation to Discord's API to do it for us, since they reserve the right to change the API at any point, for any reason, allowing us to immediately adapt to changes without having to update anything. Since, however, being developer-friendly is a major proposition of Poise and as you said, the limits are publicly documented, this can be excused, as it'll allow a developer to catch mistakes right at compile time.

PR #381, opened 2026-01-31
Check interaction type in command matching.

### LINKED ISSUE #380: Command interaction dispatch should consider command type; currently allows incorrect matches
https://github.com/serenity-rs/poise/blob/3d2a7765df66c53312a334fb26fe1846b8fa9d91/src/dispatch/slash.rs#L13-L14

Dispatching interactions for both chat input commands and user/message context commands goes through the dispatch logic in slash.rs, which matches against command name only.

[Discord provides](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-application-command-data-structure) the command's snowflake id, name, and type. (Obviously matching against the snowflake id is troublesome, since it's not known until after the command is created and will vary from guild to guild.)

However, Discord [does allow reuse of names](https://discord.com/developers/docs/interactions/application-commands#registering-a-command) as long as they have different types. You could have a "makefun" chat input (aka "slash") command, a "makefun" user context command, and a "makefun" message context command.

This is fairly unlikely, I admit, but I did encounter an issue in my bot where I had a user context command and message context command with the same name, and one of them didn't work properly because dispatch always found the other one first. I worked around it by renaming the message context command. But it would be better to check the `interaction.data.kind` value against the command being checked.

inklesspen:
This prevents errors from trying to invoke a message interaction on a user command or vice versa.

Fixes #380.

### DIFF
--- a/src/dispatch/slash.rs
+++ b/src/dispatch/slash.rs
@@ -5,6 +5,7 @@ use crate::serenity_prelude as serenity;
 /// Check if the interaction with the given name and arguments matches any framework command
 fn find_matching_command<'a, 'b, U, E>(
     interaction_name: &str,
+    interaction_kind: serenity::CommandType,
     interaction_options: &'b [serenity::ResolvedOption<'b>],
     commands: &'a [crate::Command<U, E>],
     parent_commands: &mut Vec<&'a crate::Command<U, E>>,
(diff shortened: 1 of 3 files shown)


meditationmind:
Just tested this. Was able to reproduce the error prior to fix, and verified that the fix works with a `USER`, `MESSAGE`, and `CHAT_INPUT` command all sharing the same name.

arqunis (approved):

PR #386, opened 2026-02-06
Show the error source chain in `builtins::on_error`

SabrinaJewson:
It’s more useful if the entire error source chain is shown instead of just the first error. To achieve this, we accept `E: Into<Box<dyn std::error::Error + Send + Sync>>`. Importantly, this trait is implemented for:
- any type `E: 'static + Send + Sync + Error`
- `Box<dyn std::error::Error + Send + Sync>`
- `anyhow::Error`

Because of this, in my opinion it’s not too much more onerous of a bound than `E: Debug + Display`.

### DIFF
--- a/src/builtins/mod.rs
+++ b/src/builtins/mod.rs
@@ -12,6 +12,7 @@ mod paginate;
 pub use paginate::*;
 
 use crate::{serenity::CreateAllowedMentions, serenity_prelude as serenity, CreateReply};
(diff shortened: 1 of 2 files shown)


SabrinaJewson:
Oop, CI is failing. Can someone pull the three commits from `current` onto `next`?

arqunis (approved):

PR #387, opened 2026-02-06
Support `#[string]` in slash commands

SabrinaJewson:
This allows slash command parameters to use `FromStr` instead of having to implement `SlashArgument`.

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -113,7 +113,7 @@ attributes you can use on parameters:
 - `#[string]`: Indicates that a type implements `FromStr` and should be parsed from a string argument.
 - `#[rest]`: Use the entire rest of the message for this parameter (prefix-only)
 - `#[lazy]`: Can be used on Option and Vec parameters and is equivalent to regular expressions' laziness (prefix-only)
-- `#[flag]`: Can be used on a bool parameter make it optional and default to `false`; additionally,
+- `#[flag]`: Can be used on a bool parameter to make it optional and default to `false`; additionally,
   in prefix commands only, the user can pass in the parameter name literally to set it to `true`.
     - For example with `async fn my_command(ctx: Context<'_>, #[flag] my_flag: bool)`, `~my_command` or `/my_command` would set `my_flag` to false, while `~my_command my_flag` would set `my_flag` to true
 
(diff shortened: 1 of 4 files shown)


meditationmind:
Thank you for tackling this!

Dropping `FromStr` support for slash arguments felt like a big step backward in terms of user(dev)-friendliness, so I'm happy to see it return. I've just spent a bit of time testing, and everything seems to work well.

Just a small nit regarding the documentation: Instead of adding the comment about slash behavior as a bullet point, I think it would be clearer to include the shared behavior in the initial description. For example:

> - `#[flag]`: Can be used on a `bool` parameter to make the parameter optional, defaulting to `false`, and if a prefix command, it will set the `bool` to `true` if the user typed the parameter name literally.
>    - For example, with `async fn my_command(ctx: Context<'_>, #[flag] my_flag: bool)`, `~my_command` would set `my_flag` to `false`, and `~my_command my_flag` would set `my_flag` to `true`. When used as a slash command, `my_flag` would be set to `false` unless the user sets it to `true`.

Could probably drop the last sentence if that seems too verbose, but it might be helpful to explicitly state the slash behavior in the example. (Also, that's a lot of backticks - feel free to drop some.) But again, just a small nit for clarity; I wouldn't be against merging as is.

macros/src/lib.rs
```diff
@@ -113,9 +113,9 @@ attributes you can use on parameters:
 - `#[string]`: Indicates that a type implements `FromStr` and should be parsed from a string argument.
 - `#[rest]`: Use the entire rest of the message for this parameter (prefix-only)
 - `#[lazy]`: Can be used on Option and Vec parameters and is equivalent to regular expressions' laziness (prefix-only)
-- `#[flag]`: Can be used on a bool parameter to set the bool to true if the user typed the parameter name literally
-    - For example with `async fn my_command(ctx: Context<'_>, #[flag] my_flag: bool)`, `~my_command` would set my_flag to false, and `~my_command my_flag` would set my_flag to true
-    - On slash commands, this makes the parameter optional and default to `false`.
+- `#[flag]`: Can be used on a bool parameter make it optional and default to `false`; additionally,
```
meditationmind: Missing "to" before "make it optional". Otherwise, looks good to go!

arqunis (approved):

PR #389, opened 2026-02-24
Fix breakage caused by ShardRunner refactor

nwerosama:
Since Poise uses it to generate a method for accessing the Shard latency from the Context, this PR fixes the breakage caused by https://github.com/serenity-rs/serenity/pull/3493

### DIFF
--- a/src/structs/context.rs
+++ b/src/structs/context.rs
@@ -509,7 +509,7 @@ context_methods! {
     await (ping self)
     (pub async fn ping(self) -> Option<std::time::Duration>) {
         let ctx = self.serenity_context();
-        ctx.runners.get(&ctx.shard_id)?.value().0.latency
+        ctx.runner_info.try_read()?.latency
     }
 }
 
--- a/.github/Cargo-msrv.lock
+++ b/.github/Cargo-msrv.lock
@@ -29,14 +29,14 @@ dependencies = [
  "bytestring",
  "proc-macro2",
  "quote",
- "syn 2.0.114",
+ "syn 2.0.104",
 ]
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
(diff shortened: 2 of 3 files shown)


nwerosama:
Hm, not sure how I can go about fixing CI for MSRV.

src/structs/context.rs
```diff
@@ -509,7 +509,7 @@ context_methods! {
     await (ping self)
     (pub async fn ping(self) -> Option<std::time::Duration>) {
         let ctx = self.serenity_context();
-        ctx.runners.get(&ctx.shard_id)?.value().0.latency
+        ctx.runner_info.read().latency
```
meditationmind: I think it's best to use `try_read()` here to avoid blocking or panics.

```suggestion
        ctx.runner_info.try_read()?.latency
```

meditationmind (changes requested):
To fix CI, you'll need to copy your Cargo.lock over to `./.github/Cargo-msrv.lock`.

This copy is used to avoid random dependency bumps, so if you used `cargo update` to generate the Cargo.lock you've checked in here, it might be safer to delete this one and use `cargo update serenity` instead to minimize changes.

src/structs/context.rs
nwerosama: Hm, good point.

nwerosama:
> so if you used cargo update to generate the Cargo.lock you've checked in here, it might be safer to delete this one and use cargo update serenity instead to minimize changes.

I actually ran `cargo update -p serenity` before checking it into git. Seems to update like 60 dependencies related to `serenity`, and 100+ without. Will copy the lock file over.

meditationmind:
Cool, looks good to go!

arqunis (approved):

PR #392, opened 2026-04-09
Remove the derivative proc-macro

MoistPyro:
The Derivative proc-macro is unmaintained.

this pull request removes the macro from poise, and explicitly implements Debug for several structs instead.

### DIFF
--- a/src/structs/command.rs
+++ b/src/structs/command.rs
@@ -1,24 +1,24 @@
 //! The Command struct, which stores all information about a single framework command
 
+use std::fmt::Debug;
+
 use crate::{serenity_prelude as serenity, BoxFuture};
 
 use super::{CowStr, CowVec};
 
 /// Type returned from `#[poise::command]` annotated functions, which contains all of the generated
 /// prefix and application commands
-#[derive(derivative::Derivative)]
-#[derivative(Default(bound = ""), Debug(bound = ""))]
 pub struct Command<U, E> {
     // =============
     /// Callback to execute when this command is invoked in a prefix context
-    #[derivative(Debug = "ignore")]
+    //#[derivative(Debug = "ignore")]
     pub prefix_action: Option<
         for<'a> fn(
             crate::PrefixContext<'a, U, E>,
         ) -> BoxFuture<'a, Result<(), crate::FrameworkError<'a, U, E>>>,
     >,
     /// Callback to execute when this command is invoked in a slash context
-    #[derivative(Debug = "ignore")]
+    // #[derivative(Debug = "ignore")]
     pub slash_action: Option<
         for<'a> fn(
             crate::ApplicationContext<'a, U, E>,
@@ -107,17 +107,17 @@ pub struct Command<U, E> {
     /// If true, the command may only run in NSFW channels
     pub nsfw_only: bool,
     /// Command-specific override for [`crate::FrameworkOptions::on_error`]
-    #[derivative(Debug = "ignore")]
+    // #[derivative(Debug = "ignore")]
     pub on_error: Option<fn(crate::FrameworkError<'_, U, E>) -> BoxFuture<'_, ()>>,
     /// If any of these functions returns false, this command will not be executed.
-    #[derivative(Debug = "ignore")]
+    // #[derivative(Debug = "ignore")]
     pub checks: Vec<fn(crate::Context<'_, U, E>) -> BoxFuture<'_, Result<bool, E>>>,
     /// List of parameters for this command
     ///
     /// Used for registering and parsing slash commands. Can also be used in help commands
     pub parameters: Vec<crate::CommandParameter<U, E>>,
     /// Arbitrary data, useful for storing custom metadata about your commands
-    #[derivative(Default(value = "Box::new(())"))]
+    // #[derivative(Default(value = "Box::new(())"))]
     pub custom_data: Box<dyn std::any::Any + Send + Sync>,
 
     // ============= Prefix-specific data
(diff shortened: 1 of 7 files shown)


KiloOscarSix:
Is there a reason for commenting the ignore marcos out vs deleting them?

src/structs/command.rs
```diff
@@ -1,24 +1,24 @@
...
+use std::fmt::Debug;
+
 use crate::{serenity_prelude as serenity, BoxFuture};
 
 use super::{CowStr, CowVec};
 
 /// Type returned from `#[poise::command]` annotated functions, which contains all of the generated
 /// prefix and application commands
-#[derive(derivative::Derivative)]
-#[derivative(Default(bound = ""), Debug(bound = ""))]
```
DPlayer234, line 10: This removes the `Default` impl but does not add a new one.

src/structs/command.rs
```diff
@@ -145,6 +145,49 @@ pub struct Command<U, E> {
...
+            .field("aliases", &self.aliases)
+            .field("invoke_on_edit", &self.invoke_on_edit)
+            .field("track_deletion", &self.track_deletion)
+            .field("broadcast_typing", &self.broadcast_typing)
+            .field("context_menu_name", &self.context_menu_name)
+            .field("ephemeral", &self.ephemeral)
+            .field("install_context", &self.install_context)
+            .field("interaction_context", &self.interaction_context)
+            .field("__non_exhaustive", &self.__non_exhaustive)
+            .finish()
```
DPlayer234: Might just be my opinion, but since this explicitly omits fields, this should use `.finish_non_exhaustive()` instead. Also `__non_exhaustive` could be excluded as well.
*Apparently this is just more restricted. See other comment.

src/structs/command.rs
```diff
@@ -145,6 +145,49 @@ pub struct Command<U, E> {
     pub __non_exhaustive: (),
 }
 
+// manual Debug impl to remove use of derivative proc macro
+impl<U: Debug, E: Debug> Debug for Command<U, E> {
```
DPlayer234: The previous impl was equivalent to `impl<U, E> Debug for Command<U, E>`. Are the trait bounds on `U` and `E` intentional?

src/structs/framework_options.rs
```diff
@@ -3,25 +3,23 @@
 use crate::{serenity_prelude as serenity, BoxFuture};
 
 /// Framework configuration
-#[derive(derivative::Derivative)]
-#[derivative(Debug(bound = ""))]
```
DPlayer234, line 7: This isn't reimplemented. That may be fine, but it is odd. Other options structs don't get a new one either, so I'm assuming it's at the very least intentional.

src/structs/prefix.rs
```diff
@@ -55,6 +55,24 @@ pub struct PrefixContext<'a, U, E> {
     #[doc(hidden)]
     pub __non_exhaustive: (),
 }
+
+// manual Debug impl to remove use of derivative proc macro
+impl<'a, U: Debug, E: Debug> Debug for PrefixContext<'a, U, E> {
```
DPlayer234: This impl also is now more restrictive.

src/structs/slash.rs
```diff
@@ -46,6 +46,23 @@ pub struct ApplicationContext<'a, U, E> {
     #[doc(hidden)]
     pub __non_exhaustive: (),
 }
+
+// manual Debug impl to remove use of derivative proc macro
+impl<'a, U: Debug, E: Debug> Debug for ApplicationContext<'a, U, E> {
```
DPlayer234: This impl also is now more restrictive.

src/structs/slash.rs
```diff
@@ -102,6 +117,18 @@ pub enum ContextMenuCommandAction<U, E> {
     #[doc(hidden)]
     __NonExhaustive,
 }
+
+// manual Debug impl to remove use of derivative proc macro
+impl<U: Debug, E: Debug> Debug for ContextMenuCommandAction<U, E> {
```
DPlayer234: This impl also is now more restrictive, but doesn't even use the fields it would need to be restricted for.

src/structs/slash.rs
```diff
@@ -175,6 +200,22 @@ pub struct CommandParameter<U, E> {
     pub __non_exhaustive: (),
 }
 
+// manual Debug impl to remove use of derivative proc macro
+impl<U: Debug, E: Debug> Debug for CommandParameter<U, E> {
```
DPlayer234: This impl also is now more restrictive.

DPlayer234 (commented):
Broadly a bunch of manual implementations that seem to (likely accidentally) be more restrictive than the original versions (note the `Debug(bound = "")` parts in the original derive). Also, even if only noted once, `.finish_non_exhaustive()` may be more appropriate for most of these.

MoistPyro:
Making the debug more restrictive was not my intention. I'll try removing the bounds.
(classic add stuff until the compiler stops complaining)

The comments was just me keeping them to remind myself what fields to omit.
I can write a default impl later.

src/structs/framework_error.rs
```diff
@@ -211,6 +211,112 @@ pub enum FrameworkError<'a, U, E> {
     __NonExhaustive(std::convert::Infallible),
 }
 
+// manual Debug impl to remove use of derivative proc macro
+#[allow(unused)]
+impl<'a, U: Debug, E: Debug> Debug for FrameworkError<'_, U, E> {
```
MoistPyro, line 208: this Debug restriction is actually needed. Alternativly we can ignore more fields.

arqunis:
Superseded by #408

PR #394, opened 2026-04-18
Restore Option FromStr support for prefix args

meditationmind:
Restores `Option<impl FromStr>` support for prefix command arguments using the `#[string]` attribute.

### DIFF
--- a/macros/src/command/prefix.rs
+++ b/macros/src/command/prefix.rs
@@ -177,20 +177,47 @@ fn parse_string(
     token: syn::Ident,
     parsed_rest: proc_macro2::TokenStream,
 ) -> proc_macro2::TokenStream {
-    quote::quote! {
-        match <String as ::poise::PopArgument>::pop_from(
-            &args,
-            attachment_idx,
-            serenity_ctx,
-            msg
-        ).await {
-            Ok((args, attachment_idx, token)) => {
-                match <#ty as ::std::str::FromStr>::from_str(&token) {
-                    Ok(#token) => { #parsed_rest },
-                    Err(e) => error = (e.into(), Some(token)),
-                }
-            },
-            Err(e) => error = e,
+    if let Some(ty) = unwrap_generic(ty, "Option") {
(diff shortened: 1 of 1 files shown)


NotNorom:
This has been running on prod for me for the past 3 weeks without any problems btw :D

mkrasnitski (changes requested):
Functionally this makes sense, but for clarity's sake I think separate blocks for each case would be better, so that when we figure out how to interpolate `#parsed_rest` only once for each optional argument, the changes are easier to make.

meditationmind:
Yeah, that makes sense. I split it into two blocks, and I also added similar comments to make it clear we have repetition.

meditationmind:
I think I figured out how to avoid the nested interpolation. I just pushed the changes so you can review them to see if they make sense. With eight `Option` parameters, it reduced `prefix_action` from 8483 lines to 268 lines. It seems to work, based on a quick test command. If it looks good to you, I can make the same changes to the `Option<T>` code before we merge.

mkrasnitski:
I don't think this works. The idea is that we try to parse the optional parameter, and then continue parsing everything else. If we run into an error anywhere down the line, we back up to the optional parameter and try again without parsing it. The way you've written it, if the `FromStr` parsing succeeds, we simply continue parsing, and an error later down the line will not cause parsing to roll back and try again without the parameter.

The actual solution requires the proc macro to generate a definition for a recursive function which will properly handle the depth-first approach to parsing that is required here. Either way, this should be done in a separate PR so that both pieces of `Option` parsing logic can be changed at once.

meditationmind:
Figured that was way too easy.

Reverted the changes, so the original PR should be good to go.

mkrasnitski:
CI failure is unrelated (Cargo-msrv.lock wasn't updated in most recent breakage fix). Could you rebase this to target `next` instead of `serenity-next`? It's not explicitly fixing breakage due to serenity. Other than that LGTM

meditationmind:
Done!

mkrasnitski (approved):

PR #395, opened 2026-04-19
Bump MSRV to Rust 1.95

meditationmind:
Brings MSRV in line with Serenity next and fixes CI.


meditationmind:
Was checking to make sure CI tests still pass and noticed there were lots of deprecation warnings. Took a closer look at the workflows and saw numerous issues, so I added a commit to fix those.

- Updated extremely old actions causing deprecations
- Removed some simd-json remnants
- Added cache actions
- Fixed doc scraping, which has been broken for four years because of [this](https://github.com/rust-lang/cargo/pull/10343)

I also removed `slash-trait.rs`, which was removed in [#360](https://github.com/serenity-rs/poise/pull/360/changes#diff-476b53f8e194a0b8fe07ae252c6e60670ac3daa0cd05d72ba263a3c81c8e4164) and then mistakenly [re-added by a rebase](https://github.com/serenity-rs/poise/commit/095591dd7b88f8402177e3b18488a39d23dc1797).

arqunis (approved):

PR #397, opened 2026-05-14
Ignore rustfmt::skip attribute when parsing parameter attributes

### LINKED ISSUE #396: `#[rustfmt::skip]` in command argument parameters is not ignored
## Expected Behavior
It should be possible for one to add the `#[rustfmt::skip]` attribute onto command argument parameters with the macro still expanding.

## Actual behavior
The project fails to build due to an `Unknown field` error.

```
error: Unknown field: `rustfmt::skip`
  --> sentinel-factoids/src/commands.rs:28:32
   |
28 | async fn cmd(_: Context<'_>, #[rustfmt::skip] str: String) -> Result<(), Error> {
   |                                ^^^^^^^
```

## Reproduction code
```rs
#[poise::command(slash_command)]
async fn cmd(_: Context<'_>, #[rustfmt::skip] str: String) -> Result<(), Error> {
  Ok(())
}
```

Removing the `#[rustfmt::skip]` causes that code to compile correctly.

Strokkur424:
fixes #396.

It's okay to simply ignore it in this case, due to `#[rustfmt::skip]` being a purely source-code attribute used by rustfmt. There is no reason for it to be handled by poise (i.e., by including it in the expanded source.)

### DIFF
--- a/macros/src/command/mod.rs
+++ b/macros/src/command/mod.rs
@@ -185,6 +185,11 @@ pub fn command(
         let attrs: Vec<_> = pattern
             .attrs
             .drain(..)
+            .filter(|attr| {
+                !(attr.path().segments.len() == 2
+                    && attr.path().segments[0].ident == "rustfmt"
+                    && attr.path().segments[1].ident == "skip")
+            })
             .map(|attr| darling::ast::NestedMeta::Meta(attr.meta))
             .collect();
         let attrs = <ParamArgs as darling::FromMeta>::from_list(&attrs)?;


meditationmind:
I'm not necessarily against merging this, but I'm not sure it makes sense, for the reasons I mentioned in [#396](https://github.com/serenity-rs/poise/issues/396).

1. It's nonfunctional: Yes, it gets rid of the "unknown field" error, but the `#[rustfmt::skip]` does not do anything. Rustfmt continues to function as if the attribute were not there.
2. Suppressing the error is potentially misleading: With the "unknown field" error, at least you know that `#[rustfmt::skip]` is not functioning. If we suppress the error, people could end up banging their head against a wall trying to figure out why rustfmt is continuing to function normally.
3. It's arbitrary: Why stop at `#[rustfmt::skip]`? What about `#[expect(...)]`, `#[allow(...)]`, etc.?

Strokkur424:
I am starting to suspect that I am massively misunderstanding something here. Does the `#[rustfmt::skip]` genuinely not do anything if declared as a an attribute on the parameter, different to how it would to a method not attributed to the command macro?

As in is there any functional difference between these:
```rs
// no macro!
async fn run(ctx: CommandContext<_>, #[rustfmt::skip] val: String) {}
```
```rs
#[poise::comand(slash_command)]
async fn run(ctx: CommandContext<_>, #[rustfmt::skip] val: String) {}
```

Foes rustfmt just ignore the `#[rustfmt::skip]` attribute if provided as an attribute to a param to a method expanded via macro?

I am specifically working with the assumption that rustfmt works fine here, if that's not the case (rustfmt continues formatting desipte rustfmt::skip being present), might as well close the PR.


Ignoring the other compiler attributes would make sense if this were to be merged.

meditationmind:
I've been searching, but I can't find an explicit statement regarding whether `#[rustfmt::skip]` does anything with function parameters. There is a list of [permitted built-in attributes](https://doc.rust-lang.org/stable/reference/items/functions.html#railroad-FunctionParameters), but rustfmt is a [tool attribute](https://doc.rust-lang.org/stable/reference/attributes.html#tool-attributes), so its absence there is inconclusive. However, based on a bit of experimentation just now, `#[rustfmt::skip]` on a function parameter doesn't seem to have any effect on the formatting of that parameter.

Regardless, filtering it when collecting the attributes doesn't mean it gets passed into the expanded code. Rather, it simply gets thrown out. You can verify this by expanding the code. So even if the attribute were to have an effect on parameters in a normal function signature, it would remain nonfunctional in a Poise command function even after this PR.

Strokkur424:
I mean, if it had an effect on the normal function signature, that would be enough. I don't really see how anyone would care about the formatting of an expanded macro, to my knowledge that's not being formatted by rustfmt in the first place.

If the `#[rustfmt::skip]` attribute has no effect on parameters in general, that's fine, we can close this PR then. The reason I (attempted) to use it on parameters in the first place was due to the following scenario:
```rs
#[poise::command(slash_command)]
async fn run(
	ctx: Context<_>,

	// The line is not long enough for rustfmt to put the attrib. on a new line
    #[description = "Some semi long description"] arg: String
) {}
```
Adding `#[rustfmt::skip]` to the parameter (at least that's the theory I was given after I asked for this in the Rust Discord server) should've allowed for the attributes to be on a line before without rustfmt inlining it:
```rs
#[poise::command(slash_command)]
async fn run(
	ctx: Context<_>,

	#[rustfmt::skip]
    #[description = "Some semi long description"]
    arg: String
) {}
```

At this point I don't really even know anymore whether this actually even works at all, but that was at least the theory.

meditationmind:
> Adding #[rustfmt::skip] to the parameter...

That actually does work, but it's not because of `#[rustfmt::skip]`. It's just because you have two outer attributes. You could achieve the same thing by adding `#[min_length = 0]` (or any other Poise field attribute), for example.

You could also adjust `max_width` in your `rustfmt.toml` file, but that would affect the width of all your code, so you probably don't want to do that.

PR #398, opened 2026-05-21
Fix some parsing bugs for parameters in prefix commands

mkrasnitski:
Fixes some bugs introduced by #360:
1. Parsing `#[rest] Option<T>` would pass `Option<T>` to the `ArgumentConvert` trait instead of `T`, resulting in a compiler error. 
2. Parsing `#[flag]` parameters would not consume the corresponding input token, causing parsing to fail down the line. 

Additionally this PR cleans up some miscellaneous formatting in the expanded output and moves some code around to make it more readable.

### DIFF
--- a/macros/src/command/prefix.rs
+++ b/macros/src/command/prefix.rs
@@ -318,7 +318,22 @@ pub fn generate_prefix_action(inv: &Invocation) -> Result<proc_macro2::TokenStre
         .iter()
         .enumerate()
         .map(|(idx, param)| {
-            get_modifier(param).map(|modifier| PrefixParameter {
+            let modifier = get_modifier(param)?;
+            if let Some(Modifier::Rest) = modifier {
+                if idx != inv.parameters.len() - 1 {
+                    return Err(syn::Error::new(
+                        param.span,
+                        "parameter marked `#[rest]` must come last in the argument list",
+                    ));
+                }
+                if inv.args.discard_spare_arguments {
+                    return Err(syn::Error::new(
+                        param.span,
+                        "cannot combine parameter marked `#[rest]` with `discard_spare_arguments`",
+                    ));
+                }
+            }
+            Ok(PrefixParameter {
                 idx,
                 modifier,
                 ty: param.type_.clone(),
--- a/macros/src/command/prefix.rs
+++ b/macros/src/command/prefix.rs
@@ -162,13 +162,19 @@ fn parse_flag(
     parsed_rest: proc_macro2::TokenStream,
 ) -> proc_macro2::TokenStream {
     quote::quote! {
+        let mut args = args;
+        let mut attachment_idx = attachment_idx;
         let #token = match <String as ::poise::PopArgument>::pop_from(
             args,
             attachment_idx,
             serenity_ctx,
             msg,
         ).await {
-            Ok((args, attachment_idx, #token)) if #token.eq_ignore_ascii_case(#name) => true,
+            Ok((new_args, new_attachment_idx, #token)) if #token.eq_ignore_ascii_case(#name) => {
+                args = new_args;
+                attachment_idx = new_attachment_idx;
+                true
+            },
             _ => {
                 error = (concat!("Must use either `", #name, "` or nothing as a modifier").into(), None);
                 false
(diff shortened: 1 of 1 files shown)


macros/src/command/prefix.rs
```diff
@@ -156,12 +163,12 @@ fn parse_flag(
...
         let #token = match <String as ::poise::PopArgument>::pop_from(
-            &args,
+            args,
             attachment_idx,
             serenity_ctx,
-            msg
+            msg,
         ).await {
-            Ok((args, attachment_idx, token)) if token.eq_ignore_ascii_case(#name) => true,
+            Ok((args, attachment_idx, #token)) if #token.eq_ignore_ascii_case(#name) => true,
```
meditationmind:
```suggestion
        let mut args = args.clone();
        let mut attachment_idx = attachment_idx;
        let #token = match <String as ::poise::PopArgument>::pop_from(
            args,
            attachment_idx,
            serenity_ctx,
            msg,
        ).await {
            Ok((new_args, new_attachment_idx, #token)) if #token.eq_ignore_ascii_case(#name) => {
                args = new_args;
                attachment_idx = new_attachment_idx;
                true
            }
```

This could be intended behavior, but I suspect it's a bug. Not sure if you'll want to fix it in a separate PR, but I'll mention it here just in case.

`#[flag]` functions, but it doesn't consume the argument since `#parsed_rest` falls outside of the `Ok` match arm. The `bool` gets set properly but the argument is still passed on.

meditationmind (commented):
Changes look good overall. I don't normally use prefix commands so I'm not super familiar with the modifiers, but I did some light testing and everything seems to work fine.

macros/src/command/prefix.rs
mkrasnitski: You're right, I tested it out and parsing for `#[flag]` was broken and this fixes it. I don't think the clone is necessary, but I'll add the fix to this PR.

meditationmind:
You're right, the clone was unnecessary. Just tested again without issue. LGTM.

macros/src/command/prefix.rs
meditationmind, line 349: ~~Since this hasn't been merged yet, here's one more bug fix that we could include. [Relevant discussion on Discord.](https://discord.com/channels/381880193251409931/919310428344029265/1511596930050953257)~~

Disregard.

macros/src/command/prefix.rs
```diff
@@ -257,26 +270,21 @@ fn parse_param(
             let mut args = args.clone();
```
meditationmind:
```suggestion
            let mut args = args;
```

Not part of your changes, but I think this `clone()` can be dropped.

macros/src/command/prefix.rs
```diff
@@ -55,7 +55,15 @@ fn parse_prefix_params(
         let ty = &first.ty;
         let token = format_ident!("token_{}", first.idx);
         match first.modifier {
-            Some(Modifier::Lazy) => parse_lazy(ty, token, parsed_rest)?,
+            Some(Modifier::Lazy) => {
+                let Some(ty) = unwrap_generic(ty, "Option") else {
+                    return Err(syn::Error::new(
+                        ty.span(),
+                        "can only decorate `Option<T>` with #[lazy]",
```
meditationmind, line 62: We should change the docs in [lib.rs](https://github.com/serenity-rs/poise/blob/next/macros/src/lib.rs#L115) because they still say that `#[lazy]` can be used on a `Vec<T>`.

meditationmind (commented):
Two small suggestions below. Once those are addressed, looks good to go.

macros/src/command/prefix.rs
mkrasnitski: Nothing precludes adding this feature in the future. I tried doing a git blame and I don't think `#[lazy] Vec<T>` was ever actually supported.

macros/src/command/prefix.rs
mkrasnitski: Done

meditationmind (approved):

mkrasnitski:
This wasn't necessarily a bug, but I noticed a footgun where `#[rest]` was allowed to appear anywhere in the parameter list, which would result in runtime parse errors given that `#[rest]` would consume the remainder of the input. Added a check for this which raises a compile error.

arqunis (approved):

k2helix:
Would it be possible to merge this into serenity-next branch too? It'd be highly appreciated 

Thanks!

PR #401, opened 2026-06-20
Fix exponential code generation for prefix commands with many `Option<T>` parameters

### LINKED ISSUE #376: `Option<T>` parameters in `poise::command` function cause exponential compile times
# Description

Given a command like with many `Option<T>` parameters, the `poise::command` proc-macro leads to exponentially increasing compilation times for commands with more parameters. This affect does not occur for non-optional parameters.

# Expected behavior

Functions with the `poise::command` proc macro should compile in a reasonable amount of time, or there should be an explicit upper limit on the number of parameters allowed.

# Current Behavior

Below are some data from my testing

| Test Case | `cargo check` time |
|------|--------|
| 12 `String` params | 0.22s |
| 8 `Option<bool>` params | 0.22s |
| 9 `Option<bool>` params | 6.4s |
| 10 `Option<bool>` params | 27.51s |
| 8 `Option<String>` params | 0.22s |
| 9 `Option<String>` params | 25.5s |
| 10 `Option<String>` params | 2m 57s |

# Steps to reproduce

Write a function like the following and witness `cargo check` take a very long time to complete.

See [minimum reproducible example repo](https://github.com/calebberhow/poise-command-bug-report) for full project.

```rs
#[poise::command(slash_command, prefix_command)]
pub async fn test(
    _ctx: poise::Context<'_, (), Box<dyn std::error::Error + Send + Sync>>,
    _param1: Option<String>,
    _param2: Option<String>,
    _param3: Option<String>,
    _param4: Option<String>,
    _param5: Option<String>,
    _param6: Option<String>,
    _param7: Option<String>,
    _param8: Option<String>,
    // Warning: long compile times, may stall out rust extension
    _param9: Option<String>,
    // Warning: extremely long compile times, may stall out rust extension
    // _param10: Option<String>,
    // Warning: doesn't compile on my machine, as it runs out of memory (20+ GB RAM usage).
    // _param11: Option<String>,
) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
    Ok(())
}
```

# Testing environment
Operating system: Windows 11
CPU architecture: x86-64
Cargo version: cargo 1.92.0 (344c4567c 2025-10-21)

mkrasnitski:
Closes #376

This refactors how prefix command parsing is generated to avoid exponential compile times. We avoid copy-pasting parsing logic by encapsulating it in nested function definitions and then calling said function. This keeps the code generation linear in the number of command parameters.

Given that functions cannot capture variables from outside their scope, I changed the control flow such that the final leaf node in the parse tree no longer returns the full collection of parsed parameters, but rather an empty tuple. Then, each parsing function takes the tuple of parsed input after them and prefixes it with the parameter which that function is responsible for. Errors are also propagated up to the next level rather than assigning to a top-level mutable `error` variable. This might mean that some edge case error semantics are different but I tried my best to keep errors the same.

The first two commits in this PR are purely for readability, and the behavioral changes are all in the final commit.

### DIFF
--- a/macros/src/command/prefix.rs
+++ b/macros/src/command/prefix.rs
@@ -46,46 +46,67 @@ fn get_modifier(param: &CommandParameter) -> Result<Option<Modifier>, syn::Error
 
 fn parse_prefix_params(
     params: &[PrefixParameter],
-    num_parsed: usize,
-) -> Result<proc_macro2::TokenStream, syn::Error> {
+) -> Result<(proc_macro2::TokenStream, Vec<syn::Type>), syn::Error> {
     let Some((first, rest)) = params.split_first() else {
-        // Once the input is exhausted and parsing was successful, we output the tokens we parsed.
-        let idents = (0..num_parsed)
-            .into_iter()
-            .map(|i| format_ident!("token_{}", i))
-            .collect::<Vec<_>>();
-        return Ok(quote::quote! {
-            if args.is_empty() {
-                return Ok(( #( #idents, )* ));
-            }
-        });
+        // If the input is exhausted, we output success.
+        return Ok((
+            quote::quote! {
+                if args.is_empty() {
+                    return Ok(());
+                }
+                Err((None, None))
+            },
+            vec![],
+        ));
     };
 
-    // Recursively parse the parameters after this one.
-    let parsed_rest = parse_prefix_params(rest, num_parsed + 1)?;
+    // Recursively parse the parameters after this one, and wrap the generated code in a function to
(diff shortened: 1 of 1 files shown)


meditationmind (approved):
As I mentioned before, I don't normally use prefix commands so I might have overlooked something, but I tested this with various combinations using all the modifiers and triggering all the obvious errors, and everything works properly, as far as I can tell. Line count with 10 `Option`s also looks good (`9915` before -> `1158` after).

arqunis (approved):

PR #405, opened 2026-08-09
Fix `Modal::execute_with_defaults` with `None`

meditationmind:
Currently, a default value passed to `Modal::execute_with_defaults` is wrapped in an `Option` and then unwrapped to the contained value or default, to handle both `T` and `Option<T>`. This has worked well for most cases, but when a `min_length` is specified and a default value of `None` is given, Poise passes the `None` in as `String::new()`, which results in an `Invalid Form Body` error from Discord. This PR avoids passing in a value when a default is set to `None`.


meditationmind:
Re-targeted `next` branch. Clippy failures unrelated.

arqunis (approved):

Issue #334, opened 2025-01-14
`derivative` is unmaintained RUSTSEC-2024-0388


c-git:
Hi,

Thank you very much for this crate. Not sure if you're already aware but it seems `derivative` [is unmaintained](https://github.com/mcarton/rust-derivative/issues/117). Depending on how it's used that may not be a problem in the short term but just wanted to bring it to your attention. For anyone else who is checking if an issue exists I'm going to include the output from `cargo deny` below.

```sh
warning[unmaintained]: `derivative` is unmaintained; consider using an alternative
   ┌─ /home/one/bazooka_bot/Cargo.lock:44:1
   │
44 │ derivative 2.2.0 registry+https://github.com/rust-lang/crates.io-index
   │ ---------------------------------------------------------------------- unmaintained advisory detected
   │
   = ID: RUSTSEC-2024-0388
   = Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0388
   = The [`derivative`](https://crates.io/crates/derivative) crate is no longer maintained.
     Consider using any alternative, for instance:
     - [derive_more](https://crates.io/crates/derive_more)
     - [derive-where](https://crates.io/crates/derive-where)
     - [educe](https://crates.io/crates/educe)
   = Announcement: https://github.com/mcarton/rust-derivative/issues/117
   = Solution: No safe upgrade is available!
   = derivative v2.2.0
     └── poise v0.6.1
         └── bazooka-bot v0.6.0-dev
```

c-git:
Referencing issue on repo for derivate https://github.com/mcarton/rust-derivative/issues/117

MoistPyro:
#392 
I propose just removing it.

PR #408, opened 2026-09-05
Replace `derivative` with `derive_where`

meditationmind:
Supersedes [#392](https://github.com/serenity-rs/poise/pull/392) and closes [#334](https://github.com/serenity-rs/poise/issues/334).

Replaces the unmaintained `derivative` crate with `derive_where`.

I initially intended to switch to `derive_more` since there is an open PR for deriving `Default`, but `derive_more` only supports _additional_ bounds, which made it impossible to replicate `derivative`'s `bound = ""` functionality. With `derive_where`, I was able to produce equivalent expanded code. The only difference is that `derive_where` properly indicates `non_exhaustive` (`{ , .. }`), while `derivative` did not, but this is more accurate.

Since we only used `derivative` to derive `Default` for `Command`, I just added the manual implementation for that.

Verified that all expanded code is functionally equivalent and confirmed that `Debug` output displays properly.

### DIFF
--- a/.github/Cargo-msrv.lock
+++ b/.github/Cargo-msrv.lock
@@ -293,14 +293,14 @@ dependencies = [
 ]
 
 [[package]]
-name = "derivative"
-version = "2.2.0"
+name = "derive-where"
+version = "1.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


meditationmind:
Rebased to resolve merge conflicts.

arqunis (approved):

PR #413, opened 2026-09-20
Fix possible panic when using additional prefixes

meditationmind:
Adds a check to ensure message content starts with an additional prefix before returning a value in `strip_prefix`.


meditationmind:
My bad, rushed that and put the check in the wrong spot. Fixed now.

